PR #30, opened 2020-01-28
Run compilers with -Werror

pietern:
Let's see if this works out of the box.


beauby:
Closing in favor of #89.

Issue #57, opened 2020-02-05
Support SHM ringbuffer chunking writes larger than the buffer's size


lerks:
We have by default a 2MB buffer, which is plenty for the small control messages we expect to send on it. However, to support some corner cases (like messages with millions of tensors, or the SHM transport being used as a channel using the basic channel factory) we may want to implement the ability for writes larger than the buffer's size to be split into chunks which are written separately. This should be transparent to the transport's user.

beauby:
Closing now that #82 was merged.

PR #59, opened 2020-02-05
Properly fail when attempting writes larger than ring buffer.

beauby:
Related to #57

### DIFF
--- a/tensorpipe/transport/shm/connection.cc
+++ b/tensorpipe/transport/shm/connection.cc
@@ -358,6 +358,14 @@ Connection::WriteOperation::WriteOperation(
     : ptr_(ptr), len_(len), fn_(std::move(fn)) {}
 
 void Connection::WriteOperation::handleWrite(TProducer& outbox) {
+  // Attempting to write a message larger than the ring buffer. We might want to
+  // chunk it in the future.
+  const int buf_size = outbox.getHeader().kDataPoolByteSize;
+  if (len_ > buf_size) {
+    fn_(TP_CREATE_ERROR(ShortWriteError, len_, buf_size));
+    return;
+  }
+
   ssize_t ret;
 
   // Start write transaction.
--- a/tensorpipe/transport/shm/connection.cc
+++ b/tensorpipe/transport/shm/connection.cc
@@ -337,7 +337,11 @@ void Connection::ReadOperation::handleRead(TConsumer& inbox) {
     const auto tup = inbox.readInTxWithSize<uint32_t>();
     const auto ret = std::get<0>(tup);
     const auto ptr = std::get<1>(tup);
-    TP_THROW_SYSTEM_IF(ret < 0, -ret);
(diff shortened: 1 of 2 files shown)


tensorpipe/transport/shm/connection.cc
```diff
@@ -337,7 +337,11 @@ void Connection::ReadOperation::handleRead(TConsumer& inbox) {
     const auto tup = inbox.readInTxWithSize<uint32_t>();
     const auto ret = std::get<0>(tup);
     const auto ptr = std::get<1>(tup);
-    TP_THROW_SYSTEM_IF(ret < 0, -ret);
+    if (ret < 0) {
```
pietern: There are different errors that can happen. Which one should trigger the EOF error?

tensorpipe/transport/shm/connection.cc
```diff
@@ -358,6 +362,14 @@ Connection::WriteOperation::WriteOperation(
     : ptr_(ptr), len_(len), fn_(std::move(fn)) {}
 
 void Connection::WriteOperation::handleWrite(TProducer& outbox) {
+  // Attempting to write a message larger than the ring buffer. We might want to
+  // chunk it in the future.
+  const int buf_size = outbox.getHeader().kDataPoolByteSize;
+  if (len_ > buf_size) {
+    fn_(TP_CREATE_ERROR(ShortWriteError, len_, buf_size));
+    return;
+  }
```
pietern: What happens without this check? Is there a check in the ringbuffer itself that prevents this from happening?

beauby: Without this check, the `copyToRingBuffer_` method fails [here](https://github.com/pytorch/tensorpipe/blob/master/tensorpipe/util/ringbuffer/producer.h#L201), which makes `writeInTxWithSize` fail [here](https://github.com/pytorch/tensorpipe/blob/master/tensorpipe/util/ringbuffer/producer.h#L90), which makes `handleWrite` throw `system_error` [here](https://github.com/pytorch/tensorpipe/blob/master/tensorpipe/transport/shm/connection.cc#L376).

We could check the return value of `writeInTxWithSize` but it will return `EINVAL` in different cases (write of size zero, write of size larger than buffer size, write outside of a transaction), hence the specific check here. Tehcnically, a write outside transaction should not happen unless we mess up, and a zero-write can easily be identified, so we could indeed use `EINVAL` to mean "write larger than buffer size" (if the buffer is too full, the error returned is `ENOSPC`, so we can distinguish the two).

tensorpipe/transport/shm/connection.cc
beauby: In this case we should probably only trigger `EOFError` when `-ret == ENODATA`. We could also trigger `ShortRead` when read is larger than buffer size, and throw in other cases.

beauby:
Closing in favor of #62.

PR #60, opened 2020-02-05
[don't review] Test use of ringbuffer for shm notifications

pietern:
This is terrible code intended to just get a few measurements.

### DIFF
--- a/tensorpipe/transport/shm/loop.cc
+++ b/tensorpipe/transport/shm/loop.cc
@@ -11,6 +11,7 @@
 #include <sys/eventfd.h>
 
 #include <tensorpipe/common/defs.h>
+#include <tensorpipe/util/ringbuffer/shm.h>
 
 namespace tensorpipe {
 namespace transport {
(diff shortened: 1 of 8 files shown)


tensorpipe/transport/shm/socket.h
```diff
@@ -210,13 +210,29 @@ class Socket final : public Fd, public std::enable_shared_from_this<Socket> {
   // Send file descriptor.
   template <typename... Fds>
   Error sendFds(const Fds&... fds) {
-    return sendFdsToSocket(fd_, fds...);
+    struct {
+    } nothing;
```
lerks: I was under the impression that the payload must contain at least one byte, but I cannot find any confirmation of that at the moment. Does that ring a bell? Do you know why it was one byte before?

tensorpipe/transport/shm/loop.cc
```diff
@@ -222,6 +231,40 @@ void Loop::join() {
...
   thread_.join();
+  notificationMonitor_.join();
+}
+
+void Loop::notificationMonitor() {
+  while (!done_) {
+    uint32_t token;
+    auto ret = notification_->copy(token);
+    if (-ret == ENODATA) {
+      sched_yield();
```
lerks: I was going to complain loudly about the busy waiting but this makes me kinda ok with it.

pietern: We can still do better by adding a mutex and condition variable to the ring buffer header and having producers of these notifications signal there. Then this thread can be suspended if nothing is happening.

pietern:
@lerks This is a total hack and we need a more disciplined approach :-)

pietern:
This was a hacky proof of concept. The proper version is tested at #71.

PR #68, opened 2020-02-10
Add reactor thread for shm transport

pietern:
Busy poll a ring buffer for tokens and executes std::function
associated with every incoming token. The process hosting the reactor
can share the file descriptors of its ring buffer together with a
token to let another process trigger a specific std::function.

### DIFF
--- a/tensorpipe/test/transport/shm/reactor_test.cc
+++ b/tensorpipe/test/transport/shm/reactor_test.cc
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <tensorpipe/common/defs.h>
+#include <tensorpipe/common/queue.h>
+#include <tensorpipe/transport/shm/reactor.h>
+#include <tensorpipe/transport/shm/socket.h>
+
+#include <gtest/gtest.h>
+
+using namespace tensorpipe::transport::shm;
+
+namespace {
+
+void run(std::function<void(int)> fn1, std::function<void(int)> fn2) {
+  int fds[2];
+
+  {
+    auto rv = socketpair(AF_UNIX, SOCK_STREAM, 0, fds);
+    if (rv != 0) {
+      TP_THROW_SYSTEM(errno) << "Failed to create socket pair";
+    }
+  }
+
+  {
+    auto pid = fork();
+    TP_DCHECK_GE(pid, 0);
+    if (pid == 0) {
+      close(fds[0]);
+      fn2(fds[1]);
+      close(fds[1]);
+      exit(0);
+    }
+  }
+
+  close(fds[1]);
+  fn1(fds[0]);
+  close(fds[0]);
+  wait(nullptr);
+}
+
+} // namespace
+
+TEST(Reactor, Basic) {
+  run(
+      [](int fd) {
+        tensorpipe::Queue<int> queue;
+        auto reactor = std::make_shared<Reactor>();
+        auto token = reactor->add([&] { queue.push(1); });
+
+        // Share reactor fds and token with other process.
+        {
+          auto socket = Socket(fd);
+          auto fds = reactor->fds();
+          auto error = socket.sendPayloadAndFds(
+              token, std::get<0>(fds), std::get<1>(fds));
+          ASSERT_FALSE(error) << error.what();
+        }
+
+        // Wait for other process to run trigger.
+        queue.pop();
+      },
+      [](int fd) {
+        Reactor::TToken token;
+        Fd header;
+        Fd data;
+
+        // Wait for other process to share reactor fds and token.
+        {
+          auto socket = Socket(fd);
+          auto error = socket.recvPayloadAndFds(token, header, data);
+          ASSERT_FALSE(error) << error.what();
+        }
+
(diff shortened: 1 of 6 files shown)


tensorpipe/transport/shm/reactor.h
```diff
@@ -0,0 +1,113 @@
...
+#include <tensorpipe/util/ringbuffer/consumer.h>
+#include <tensorpipe/util/ringbuffer/producer.h>
+
+namespace tensorpipe {
+namespace transport {
+namespace shm {
+
+// Extra data stored in ring buffer header.
+struct ReactorRingBufferExtraData {
+  pthread_mutex_t mutex_;
```
lerks: Is this used? If not, please remove it.

tensorpipe/transport/shm/reactor.h
```diff
@@ -0,0 +1,113 @@
...
+  std::atomic_bool done_{false};
+
+  // Reactor thread entry point.
+  void run();
+
+  // Next token to give out if there are no reusable ones.
+  TToken nextToken_{0};
+
+  // Tokens are placed in this set if they can be reused.
+  std::set<TToken> reusableTokens_;
```
lerks: Not sure about the complexity guarantees, but if all you do is add one and remove the smallest, you can use a heap backed by a vector? The STL has algorithms to deal with heaps easily.

tensorpipe/transport/shm/reactor.h
```diff
@@ -0,0 +1,113 @@
...
+
+  std::mutex mutex_;
+  std::thread thread_;
+  std::atomic_bool done_{false};
+
+  // Reactor thread entry point.
+  void run();
+
+  // Next token to give out if there are no reusable ones.
+  TToken nextToken_{0};
```
lerks: You don't even really need this, its value will always be equal to `functions_.size() + reusableTokens_.size()`.

tensorpipe/transport/shm/reactor.cc
```diff
@@ -0,0 +1,121 @@
...
+  std::unique_lock<std::mutex> lock(mutex_);
+  functions_[token] = nullptr;
+  reusableTokens_.insert(token);
+}
+
+void Reactor::trigger(TToken token) {
+  for (;;) {
+    auto rv = producer_->write(token);
+    if (rv == -EAGAIN) {
+      continue;
```
lerks: Uhm, perhaps yield here?

tensorpipe/transport/shm/reactor.cc
```diff
@@ -0,0 +1,121 @@
...
+
+std::tuple<int, int> Reactor::fds() const {
+  return std::make_tuple(headerFd_.fd(), dataFd_.fd());
+}
+
+void Reactor::run() {
+  while (!done_.load()) {
+    uint32_t token;
+    auto ret = consumer_->copy(token);
+    if (-ret == ENODATA) {
```
lerks: Nit: `ret == -ENODATA`. It's consistent with everything else and I believe this way the compiler can do the negation at compile time rather than at runtime.

tensorpipe/transport/shm/reactor.cc
```diff
@@ -0,0 +1,121 @@
...
+      sched_yield();
+      continue;
+    }
+
+    TFunction fn;
+
+    // Make copy of std::function so we don't need
```
lerks: Ugh, this will not be nice to handle when we switch to non-copyable functions...

tensorpipe/transport/shm/reactor.cc
```diff
@@ -0,0 +1,121 @@
...
+    : producer_(util::ringbuffer::shm::load<TReactorRingBuffer>(
+          headerFd.fd(),
+          dataFd.fd())),
+      token_(token) {}
+
+void Reactor::Trigger::run() {
+  for (;;) {
+    auto rv = producer_.write(token_);
+    if (rv == -EAGAIN) {
+      continue;
```
lerks: Yield?

tensorpipe/transport/shm/reactor.cc
```diff
@@ -0,0 +1,121 @@
...
+    }
+    TP_DCHECK_EQ(rv, sizeof(token_));
+    break;
+  }
+}
```
lerks: This function does the same thing as Reactor::trigger, is there a way to remove one or have one call the other?

tensorpipe/transport/shm/reactor.cc
```diff
@@ -0,0 +1,121 @@
...
+    if (fn) {
+      fn();
+    }
+  }
+}
+
+Reactor::Trigger::Trigger(Fd&& headerFd, Fd&& dataFd, TToken token)
+    : producer_(util::ringbuffer::shm::load<TReactorRingBuffer>(
+          headerFd.fd(),
+          dataFd.fd())),
```
lerks: I'm surprised this doesn't fail: both the Fd's and the Segment's destructors will close the fd, which should cause the second close to fail, or to do something quite bad.

lerks: Since you're moving them into here (they're rvalue refs), however, it means the Fds are now ours, so perhaps you can call `.release()` on them?

tensorpipe/test/transport/shm/reactor_test.cc
```diff
@@ -0,0 +1,111 @@
...
+    if (pid == 0) {
+      close(fds[0]);
+      fn2(fds[1]);
+      exit(0);
+    }
+  }
+
+  close(fds[1]);
+  fn1(fds[0]);
+  wait(nullptr);
```
lerks: And then close `fds[0]` too?

PR #71, opened 2020-02-10
[wip] Reactor integration

pietern:
@lerks Can you check out commits 2 and 3. I'll submit them as separate PRs to retain history.

### DIFF
--- a/tensorpipe/transport/shm/loop.cc
+++ b/tensorpipe/transport/shm/loop.cc
@@ -100,6 +100,10 @@ void Loop::defer(TDeferredFunction fn) {
   reactor_->trigger(deferredFunctionReactorToken_);
 }
 
+const std::shared_ptr<Reactor>& Loop::reactor() {
+  return reactor_;
+}
+
 void Loop::registerDescriptor(
     int fd,
     int events,
--- a/tensorpipe/transport/shm/loop.cc
+++ b/tensorpipe/transport/shm/loop.cc
@@ -76,6 +76,8 @@ Loop::Loop(ConstructorToken /* unused */) {
   reactor_ = std::make_shared<Reactor>();
   epollReactorToken_ =
       reactor_->add([this] { handleEpollEventsFromReactor(); });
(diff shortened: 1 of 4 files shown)


tensorpipe/transport/shm/loop.cc
```diff
@@ -224,6 +197,58 @@ void Loop::join() {
   thread_.join();
 }
 
+void Loop::handleEpollEventsFromReactor() {
+  std::unique_lock<std::mutex> lock(epollMutex_);
+
+  // Process events returned by epoll_wait(2).
+  {
+    std::unique_lock<std::mutex> lock(handlersMutex_);
```
lerks: I am slightly worried by the fact that here (and later with `mutex_`) we're acquiring a lock while holding another. This could result in a deadlock if someone else is trying to do the same in reverse order (I don't think it currently happens, but you never know). I would be more comfortable if we used something like scoped_lock, that can acquire multiple locks in a "dealock-avoiding" fashion (if acquiring one fails, it releases all of them before retrying). We could use it here, except that then we might briefly give up the `epollMutex_`. I don't think that's a problem, as if we can acquire it once it means the epoll loop is waiting for the condition and thus won't try to reacquire it until we set the condition. This is all quite complicated...

lerks: In fact, don't we only need to hold `epollMutex_` inside this scope, and can release it in the next scope? We can have the epoll loop resume while we process handlers, that shouldn't hurt...

pietern: The `epollMutex_` and condition variable allow us to ping pong between the epoll thread and the reactor thread. The other lock is used to provide atomic access to the handler list. They are separate and they won't be used in reverse order. In the unlikely case that sometime in the future, they do, I anticipate tsan will step in...,

pietern:
These commits will be merged separately to retain the history.

pietern:
This is superseded by the ghstack PRs at #74, #75, and #76.

PR #81, opened 2020-02-14
Implement write queueing for SHM.

beauby:
Summary:
Leverage reactors and triggers to signal a peer that its outbox has been read from, and hence has more available space for potential subsequent writes.
This involves each peer to have two reactor tokens, that both need to be exchanged.

Differential Revision: D19884014

### DIFF
--- a/tensorpipe/test/transport/shm/connection_test.cc
+++ b/tensorpipe/test/transport/shm/connection_test.cc
@@ -15,7 +15,8 @@ using namespace tensorpipe::transport;
 
 using SHMConnectionTest = ConnectionTest<SHMConnectionTestHelper>;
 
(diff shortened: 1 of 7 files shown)


pietern:
I ran our benchmark with this change to test the impact. It's not a lot.

Before: 

```
address = shm://foobar
io_num = 500000
chunk_bytes = 8
chunk-size      # ping-pong     avg (usec)   p50     p75     p90     p95
8               500000          1.821        1.741   1.863   2.010   2.064
```

After:

```
address = shm://foobar
io_num = 500000
chunk_bytes = 8
chunk-size      # ping-pong     avg (usec)   p50     p75     p90     p95
8               500000          1.900        1.864   1.981   2.085   2.135
```

PR #89, opened 2020-02-21
Run compilers with -Werror

beauby:
Commandeering @pietern's #30.

This currently blocks because of deprecated googletest macros.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,8 @@ jobs:
             mkdir build
             cd build
             cmake ../ \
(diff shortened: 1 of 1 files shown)


lerks:
It would probably also block on some deprecated protobuf functions: we need to support v3.0, which only has ByteSize(), but on my personal machine I have the latest version where they have introduced ByteSizeLong() and deprecated ByteSize(). We can't have -Werror if we want to support both v3.0 and the latest one. (We must support v3.0 because it's the one in Ubuntu 18.04 I think)

Is there a way to have -Werror _except_ deprecation warnings?

PR #112, opened 2020-03-09
Bundle the Python bindings in a single shared library

### LINKED ISSUE #110: Bundle Python bindings in single shared library
Currently when building the Python bindings we end up with a lot of separate shared libraries:
```
$ ls venv/lib/python3.6/site-packages/
easy-install.pth                              libtensorpipe_uv.so                           pkg_resources-0.0.0.dist-info/
easy_install.py                               libuv.so                                      __pycache__/
libtensorpipe_basic.so                        libuv.so.1                                    pytensorpipe.cpython-36m-x86_64-linux-gnu.so
libtensorpipe_proto_channel.so                libuv.so.1.0.0                                setuptools/
libtensorpipe_proto.so                        pip/                                          setuptools-39.0.1.dist-info/
libtensorpipe_shm.so                          pip-9.0.1.dist-info/                          tensorpipe-0.0.0.egg-info/
libtensorpipe.so                              pkg_resources/
```
Basically, it seems our internal CMake target structure is replicated as .so files.

This has already proven to cause problems. For example, I'm getting this:
```
>>> import pytensorpipe
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: libtensorpipe_basic.so: cannot open shared object file: No such file or directory
```

I'm not sure what's the exact cause of that problem, but I suspect it may be related to the use of separate .so files. Perhaps we should try to stick all of TensorPipe (+libuv) into a single .so file.

lw:
Except libuv, as their CMakeLists.txt file enforces building it as a shared library.

Fixes #110.

### DIFF
--- a/tensorpipe/CMakeLists.txt
+++ b/tensorpipe/CMakeLists.txt
@@ -52,8 +52,6 @@ add_library(tensorpipe ${TENSORPIPE_SOURCES})
 target_include_directories(tensorpipe PUBLIC ${CMAKE_SOURCE_DIR})
 target_link_libraries(tensorpipe PUBLIC tensorpipe_proto)
 
-target_link_libraries(tensorpipe PUBLIC uv)
-
 add_subdirectory(channel/basic)
 add_subdirectory(transport/shm)
 add_subdirectory(transport/uv)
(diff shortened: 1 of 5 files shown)


tensorpipe/CMakeLists.txt
```diff
@@ -58,9 +58,7 @@ add_subdirectory(channel/basic)
 add_subdirectory(transport/shm)
 add_subdirectory(transport/uv)
 
-if(BUILD_PYTHON_MODULE)
```
beauby: Why remove the guard option?

PR #116, opened 2020-03-10
Streamline cmake architecture

beauby:
Depends on #115.

Before:
graph

After:
graph


lw:
CI is broken.

Also, I suspect you included more commits in this than you intended.

I'm not sure the extreme simplification of the new structure is always desirable. I didn't mind having a separate target for each transport/channel and the modularity that it brought, especially as it mirrored the modularity of the code. With the new organization it's a all or nothing. For example, if I only want the Python bindings, I also need to build all the benchmarks.

At the moment I'm a bit more inclined to stick to the old structure, but it would help if you could clarify what the advantages of the new approach are.

beauby:
graph

Reworked this one. There is now only one tensorpipe library (+ the proto one for now), instead of one for each transport/channel + the core library. This will simplify packaging and linking for end users, as they can simply link to `tensorpipe`. Moreover, all tests are built in a single binary.

PR #119, opened 2020-03-11
Fix listener queueing tests

beauby:
These tests were hanging on OSX because the `Connection`s were destroyed before the call to `uv_tcp_connect()` fired its callback.


beauby:
The 2 failures are unrelated:
```
The following tests FAILED:
	 29 - Shm/TransportTest.Connection_InitializationError/0x5557ce76d018 (Failed)
	 31 - Shm/TransportTest.Connection_ProtobufWrite/0x5557ce76d018 (Child aborted)
```

PR #133, opened 2020-03-23
Introduce config.h.

beauby:
Introducing a header that centralizes the `#define`s for various optional parts (transports/channels), mapping directly from CMake options.


beauby:
Closing as the functionality has been implemented since.

PR #176, opened 2020-06-04
Channel/transport test refactor for process based tests.

beauby:
Introduce `PeerGroup` class, along with its subclasses `ThreadPeerGroup` and `ProcessPeerGroup`. Those abstract spawning/joining a group of peers, as well as communication and synchronization among them.

The rationale behind this is to pave the way for testing CUDA channels.

### DIFF
--- a/tensorpipe/test/channel/channel_test.h
+++ b/tensorpipe/test/channel/channel_test.h
@@ -10,57 +10,75 @@
 
 #include <future>
 #include <memory>
+#include <string>
 #include <thread>
 
 #include <tensorpipe/channel/context.h>
-#include <tensorpipe/common/queue.h>
 #include <tensorpipe/transport/uv/context.h>
 
 #include <gtest/gtest.h>
 
+#include <tensorpipe/test/peer_group.h>
+
 class ChannelTestHelper {
  public:
+  virtual ~ChannelTestHelper() = default;
+
   virtual std::shared_ptr<tensorpipe::channel::Context> makeContext(
       std::string id) = 0;
 
-  virtual ~ChannelTestHelper() = default;
+  virtual std::shared_ptr<PeerGroup> makePeerGroup() {
+    return std::make_shared<ThreadPeerGroup>();
+  }
 };
 
 class ChannelTest : public ::testing::TestWithParam<ChannelTestHelper*> {
+ protected:
+  std::shared_ptr<PeerGroup> peers_;
+
  public:
-  void testConnectionPair(
+  ChannelTest() : peers_(GetParam()->makePeerGroup()) {}
+
+  void testConnection(
       std::function<void(std::shared_ptr<tensorpipe::transport::Connection>)>
-          f1,
+          server,
       std::function<void(std::shared_ptr<tensorpipe::transport::Connection>)>
-          f2) {
-    auto context = std::make_shared<tensorpipe::transport::uv::Context>();
-    context->setId("harness");
+          client) {
     auto addr = "127.0.0.1";
 
-    {
-      tensorpipe::Queue<std::shared_ptr<tensorpipe::transport::Connection>> q1,
-          q2;
+    peers_->spawn(
+        [&] {
+          auto context = std::make_shared<tensorpipe::transport::uv::Context>();
+          context->setId("server_harness");
 
-      // Listening side.
-      auto listener = context->listen(addr);
-      listener->accept(
-          [&](const tensorpipe::Error& error,
-              std::shared_ptr<tensorpipe::transport::Connection> connection) {
-            ASSERT_FALSE(error) << error.what();
-            q1.push(std::move(connection));
-          });
+          auto listener = context->listen(addr);
 
-      // Connecting side.
-      q2.push(context->connect(listener->addr()));
+          std::promise<std::shared_ptr<tensorpipe::transport::Connection>>
+              connectionProm;
+          std::future<std::shared_ptr<tensorpipe::transport::Connection>>
+              connectionFuture = connectionProm.get_future();
+          listener->accept(
+              [&](const tensorpipe::Error& error,
+                  std::shared_ptr<tensorpipe::transport::Connection>
+                      connection) {
+                ASSERT_FALSE(error) << error.what();
+                connectionProm.set_value(std::move(connection));
+              });
 
-      // Run user specified functions.
-      std::thread t1([&] { f1(q1.pop()); });
-      std::thread t2([&] { f2(q2.pop()); });
-      t1.join();
-      t2.join();
-    }
+          peers_->send(PeerGroup::kClient, listener->addr());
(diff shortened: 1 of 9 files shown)


beauby:
There seems to be a data race in `uv_loop_init()`, but might be a false positive.

tensorpipe/test/peer_group.h
```diff
@@ -0,0 +1,136 @@
...
+
+    return str;
+  }
+
+  void spawn(std::function<void()> f1, std::function<void()> f2) override {
+    std::function<void()> fns[] = {std::move(f1), std::move(f2)};
+    pid_t pids[] = {-1, -1};
+
+    for (int i = 0; i < 2; ++i) {
+      pids[i] = fork();
```
lw, line 117: After we fork do we perhaps want to close some of the pipes? The way this currently works, is that all processes (parent, child0 and child1) all keep all the pipes open. If now child1 dies unexpectedly, the parent won't notice (because it is first waiting for child0 to finish, and it is still alive), and child0 may be blocked on a read from a pipe from child1. That read will never return because no one will ever write on that pipe. The read won't fail either, because both child0 and parent still have that pipe open and thus could potentially write to it. Thus we're in a deadlock.

A solution to the above problem is for the parent to immediately close all the pipes, and for each child to close one endpoint of each pipe. This way each endpoint is only open in exactly one process. That way, if that process dies unexpectedly, it will also in turn close one end of the pipe (no one else is keeping it alive) and thus anyone reading from the other end would correctly get an error.

tensorpipe/test/peer_group.h
```diff
@@ -0,0 +1,136 @@
...
+
+  void sendPeer(int peerId, const std::string& str) override {
+    write(pipefd[peerId][1], str.c_str(), str.length());
+    write(pipefd[peerId][1], "\n", 1);
+  }
+
+  std::string recvPeer(int peerId) override {
+    std::string str;
+    char c;
+    while (read(pipefd[peerId][0], &c, 1) > 0) {
```
lw, line 102: It probably makes sense to check the return code of the `read` and `write` calls and if they signal an error print a nice readable message. And, in fact, we should probably also do it for `pipe` (above) and `fork` (below).

tensorpipe/test/transport/transport_test.h
```diff
@@ -19,51 +18,69 @@
...
-      // Wait for completion.
-      listeningThread.join();
-      connectingThread.join();
-    }
+    peers_->spawn(
+        [&] {
+          auto ctx = GetParam()->getContext();
+          auto addr = GetParam()->defaultAddr();
+          auto listener = ctx->listen(addr);
+          tensorpipe::Queue<std::shared_ptr<Connection>> queue;
```
lw, line 55: Nit: this queue could be replaced by a promise

tensorpipe/test/peer_group.h
```diff
@@ -0,0 +1,136 @@
...
+    cond_var_[peerId].notify_one();
+  }
+
+  void joinPeer(int peerId) {
+    while (recvPeer(peerId) != kDone)
+      ;
+    std::unique_lock<std::mutex> lock(m_);
+    while (!done_[peerId]) {
+      cond_var_[peerId].wait(lock);
+    }
```
lw, line 41:
```suggestion
    cond_var_[peerId].wait(lock, [&]() { return done_[peerId]; });
```

tensorpipe/test/transport/shm/connection_test.cc
```diff
@@ -107,57 +94,44 @@ TEST_P(ShmTransportTest, QueueWrites) {
...
-  std::promise<void> writeScheduledProm;
-  std::promise<void> writeCompletedProm;
-  std::promise<void> readCompletedProm;
-  std::future<void> writeCompletedFuture = writeCompletedProm.get_future();
-  std::future<void> readCompletedFuture = readCompletedProm.get_future();
 
-  this->testConnection(
+  testConnection(
       [&](std::shared_ptr<Connection> conn) {
-        writeScheduledProm.get_future().get();
```
lw, line 118: This `writeScheduledProm` has been removed but hasn't been replaced with anything. Based on the test's name, the reason it was there was to ensure the writes had all been scheduled and queued before starting to do the reads.

tensorpipe/test/channel/channel_test.h
```diff
@@ -18,49 +19,81 @@
...
-      std::thread t2([&] { f2(q2.pop()); });
-      t1.join();
-      t2.join();
-    }
-
-    context->join();
+    peers_->spawn(
+        [&] {
+          auto context = std::make_shared<tensorpipe::transport::uv::Context>();
+          context->setId("harness");
```
lw, line 53: As you're now creating two separate contexts in the client and server, could you give them different IDs? Like, `client_harness` and `server_harness`. Otherwise we won't be able to tell them apart in the logs

tensorpipe/test/channel/channel_test.h
```diff
@@ -18,49 +19,81 @@
...
+
+          auto listener = context->listen(addr);
+
+          tensorpipe::Queue<std::shared_ptr<tensorpipe::transport::Connection>>
```
lw, line 57: This queue could be a promise

tensorpipe/test/peer_group.h
```diff
@@ -0,0 +1,136 @@
...
+    std::function<void()> fns[] = {std::move(f1), std::move(f2)};
+    pid_t pids[] = {-1, -1};
+
+    for (int i = 0; i < 2; ++i) {
+      pids[i] = fork();
+      if (pids[i] == 0) {
+        fns[i]();
+        exit(testing::Test::HasFailure());
+        return;
+      }
```
lw, line 122: Could we add a `else` branch where we print the PID of the children? It would help if one needs to attach GDB to them...

lw:
BTW, yes, there really seems to be a race in libuv. This function is reading/writing to a global static variable without a mutex:
https://github.com/libuv/libuv/blob/37a8affbd319209fd0d0d96ee4e3002c8ecc5076/src/unix/linux-core.c#L473-L503

We could report it to them, but I suspect they will tell us that you're not supposed to create two libuv loops in the same process. Also, even if they fix it, it will only become available in the next version. Do you have any idea of how we could avoid that TSAN error? Using processes instead of threads would work, but perhaps there are other workarounds?

tensorpipe/test/peer_group.h
beauby: We could, but we could also go further and tag all logs with the PID. What do you think?

tensorpipe/test/transport/shm/connection_test.cc
beauby: The purpose of the test (making sure writes are properly queued) is ensured implicitly by having payloads larger than half of the ringbuffer size, but I'll add some message passing to make it explicit.

PR #196, opened 2020-08-04
Gate more optional features behind CMake flags

lw:
* #198 Allow to build TensorPipe as a static lib
* #197 Allow downstream projects to customize install location
* **#196 Gate more optional features behind CMake flags**

The idea here is that most people who build TensorPipe do it for the library itself, it's just us developers that need to build tests, or benchmarks, etc. So why not leave them disabled by default, so that there's no risk of users getting it wrong?

Differential Revision: [D22928853](https://our.internmc.facebook.com/intern/diff/D22928853)

### DIFF
--- a/tensorpipe/CMakeLists.txt
+++ b/tensorpipe/CMakeLists.txt
@@ -134,12 +134,14 @@ endif()
 
 ## Benchmarks
 
-add_subdirectory(benchmark)
+if (TP_BUILD_BENCHMARK)
+  add_subdirectory(benchmark)
+endif()
 
 
 ## Tests
 
-if(BUILD_TESTING)
+if(TP_BUILD_TESTING)
   add_subdirectory(test)
 endif()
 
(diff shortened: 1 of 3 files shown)


tensorpipe/CMakeLists.txt
```diff
@@ -134,12 +134,14 @@ endif()
...
-add_subdirectory(benchmark)
+if (TP_BUILD_BENCHMARK)
+  add_subdirectory(benchmark)
+endif()
 
 
 ## Tests
 
-if(BUILD_TESTING)
+if(TP_BUILD_TESTING)
```
beauby, line 144: `BUILD_TESTING` is the standard CTest variable. We're currently not using CTest anymore, but if we do switch back to it down the line, we'll want to use that variable again, and it might break builds that depend on tensorpipe by `include_directory()`.

lw: Yeah, I saw that BUILD_TESTING is "standard", but it doesn't give us any benefit and in fact hurts downstream projects because they need to temporarily set that variable and then possibly revert it if they are using it themselves. This was the case with PyTorch (where, in fact, we weren't even resetting it correctly because we had a typo).

beauby: Then maybe we could at least make it default to the value of `BUILD_TESTING` when `TP_BUILD_TESTING` is not set?

PR #197, opened 2020-08-04
Allow downstream projects to customize install location

lw:
* #198 Allow to build TensorPipe as a static lib
* **#197 Allow downstream projects to customize install location**
* #196 Gate more optional features behind CMake flags

It makes perfect sense for us to install in the standard location for libraries in the system. However, this sometimes turns out to be `lib64`, while PyTorch hardcodes such a location as `lib`. Not sure it matters, and if anything it should be PyTorch to fix this, but let's be nice and allow users to specify a different location if they want.

Differential Revision: [D22928855](https://our.internmc.facebook.com/intern/diff/D22928855)


cmake/Options.cmake
```diff
@@ -23,3 +23,8 @@ option(TP_BUILD_TESTING "Build tests" OFF)
 
 # Force to build libuv from the included submodule
 option(TP_BUILD_LIBUV "Build libuv from source" OFF)
+
+# Directories
+include(GNUInstallDirs)
+set(TP_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR} CACHE STRING "Directory in which to install libraries")
```
malfet, line 29: To make this essentially a no-op shouldn't it be set by default to `${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}`
```suggestion
set(TP_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR} CACHE STRING "Directory in which to install libraries")
```

PR #198, opened 2020-08-04
Allow to build TensorPipe as a static lib

lw:
* **#198 Allow to build TensorPipe as a static lib**
* #197 Allow downstream projects to customize install location
* #196 Gate more optional features behind CMake flags

Sometimes (I'm looking at you, PyTorch) users include TensorPipe as a submodule in a project, where they want to build the top-level library as shared, and thus BUILD_SHARED_LIBS is enabled, but they want to build and link TensorPipe *statically* into that shared library. So we should offer a way for users to override the default behavior. (However, even when TensorPipe is built statically, if the final result is dynamic we must build position-independent code).

Differential Revision: [D22928857](https://our.internmc.facebook.com/intern/diff/D22928857)

### DIFF
--- a/tensorpipe/CMakeLists.txt
+++ b/tensorpipe/CMakeLists.txt
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 add_library(tensorpipe
+  ${TP_STATIC_OR_SHARED}
   channel/error.cc
   channel/helpers.cc
   channel/registry.cc
@@ -118,6 +119,11 @@ endif()
 
 target_link_libraries(tensorpipe PRIVATE protobuf::libprotobuf)
 protobuf_generate(TARGET tensorpipe)
+
+
+if(BUILD_SHARED_LIBS)
+  set_target_properties(tensorpipe PROPERTIES POSITION_INDEPENDENT_CODE 1)
+endif()
 
 
 ## Config
--- a/cmake/Options.cmake
+++ b/cmake/Options.cmake
@@ -21,6 +21,14 @@ option(TP_BUILD_BENCHMARK "Build benchmarks" OFF)
 option(TP_BUILD_PYTHON "Build python bindings" OFF)
 option(TP_BUILD_TESTING "Build tests" OFF)
 
+# Whether to build a static or shared library
+if(BUILD_SHARED_LIBS)
+  set(TP_STATIC_OR_SHARED SHARED CACHE STRING "")
+else()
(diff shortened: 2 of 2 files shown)


tensorpipe/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 add_library(tensorpipe
+  ${TP_STATIC_OR_SHARED}
```
beauby, line 8: From the CMake docs:
> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON.

Maybe I am missing something, but this change does not seem to add anything?

tensorpipe/CMakeLists.txt
```diff
@@ -120,6 +121,11 @@ target_link_libraries(tensorpipe PRIVATE protobuf::libprotobuf)
 protobuf_generate(TARGET tensorpipe)
 
 
+if(BUILD_SHARED_LIBS)
+  set_target_properties(tensorpipe PROPERTIES POSITION_INDEPENDENT_CODE 1)
+endif()
```
beauby, line 126: Also from the CMake docs:
> For SHARED and MODULE libraries the POSITION_INDEPENDENT_CODE target property is set to ON automatically.

What we might want to enable though is for the user to say "I want a static lib, but with PIC".

tensorpipe/CMakeLists.txt
lw: Indeed, *by default* this keeps the exact same behavior as before. But it allows downstream users (PyTorch), who may want to build their library as a dynamic library, to change how TensorPipe is built without changing that "global" flag.

tensorpipe/CMakeLists.txt
lw: Yes, this is the gist. Note that this is in fact exactly what we already do in our own target for libuv (i.e., `_uv_a`), so that when TensorPipe is built dynamically we still build libuv statically but with PIC so that it can be linked against the dynamic libtensorpipe.

beauby: But here tensorpipe will only get build with PIC if `BUILD_SHARED_LIBS` is `ON`?

lw: Correct: if the global BUILD_SHARED_LIBS, which is possibly coming from a parent project, is ON, then TensorPipe will be built with PIC, even when TensorPipe is a static library due to TP_STATIC_OR_SHARED being set to STATIC by the parent project.

PR #204, opened 2020-08-30
Fix error: ‘runtime_error’ is not a member of ‘std’

cdluminate:


beauby:
Thanks for the fix @cdluminate. Would you mind fixing the clang-format warning before we merge? (Basically swapping the order of includes.)

cdluminate:
@beauby Thanks for the notification. Formated and force-pushed.

PR #205, opened 2020-09-01
Channel hierarchies.

beauby:


tensorpipe/channel/channel.cc
```diff
@@ -0,0 +1,36 @@
...
+namespace tensorpipe {
+namespace channel {
+template <>
+void Channel<CpuTensor>::send(
+    const void* ptr,
+    size_t length,
+    TDescriptorCallback descriptorCallback,
+    TSendCallback callback) {
+  send(
+      CpuTensor{const_cast<void*>(ptr), length},
```
lw, line 21: I see why we need the const_cast here, but it looks like a code smell. In other words, by introducing these CpuTensor structs, are we giving up some const-correctness? Because a const CpuTensor means that the address of its pointers is const, but it doesn't enforce constness on the memory they point to. I'm wondering if there is some semi-standard way to deal with this? If not, well, too bad, let's go ahead anyways...

tensorpipe/channel/context.h
```diff
@@ -25,6 +25,7 @@ namespace channel {
 // context. All registered instances are assumed to be eligible
 // channels for all pairs.
 //
+template <typename TTensor>
 class Context {
```
lw, line 29: As I said elsewhere, I think it would be nice if we could add this at the bottom of this file:
```
using CpuContext = Context<CpuTensor>;
using CudaContext = Context<CudaTensor>;
```
as shorthands since we'll be using them often.

tensorpipe/channel/channel.h
```diff
@@ -45,29 +45,40 @@
...
+enum class Endpoint : bool { kConnect, kListen };
+
+using TDescriptor = std::string;
+using TDescriptorCallback = std::function<void(const Error&, TDescriptor)>;
+using TSendCallback = std::function<void(const Error&)>;
+using TRecvCallback = std::function<void(const Error&)>;
+
 // Abstract base class for channel classes.
+template <typename TTensor>
 class Channel {
```
lw, line 57: Might be convenient to have
```
using CpuChannel = Channel<CpuTensor>;
using CudaChannel = Channel<CudaTensor>;
```
defined at the end of this file.

tensorpipe/common/tensor.h
```diff
@@ -0,0 +1,53 @@
...
+namespace tensorpipe {
+
+enum class DeviceType {
+  kCpu = 0,
+#if TENSORPIPE_HAS_CUDA_IPC_CHANNEL
+  kCuda,
+#endif // TENSORPIPE_HAS_CUDA_IPC_CHANNEL
+};
+
+struct CpuTensor {
```
lw, line 24: I may be overthinking this, but should we assume that these structs are always copiable (and always will be, even for future device types)? Or should we consider that some may be movable-only and thus use std::move(...) whenever we pass them around? For now I don't really see any case in which moving would be required, so I guess we shouldn't worry about this right now...

tensorpipe/common/tensor.h
```diff
@@ -0,0 +1,53 @@
...
+  void* data{nullptr};
```
lw, line 25: Mega-nit: in some cases we called this `ptr`. Not sure whether I prefer that or `data`. Maybe `ptr` conveys more the idea that this struct doesn't hold the data by itself (it doesn't store a `string`, a `vector`, ... which owns the data) but is just a shallow copyable reference to some data that exists elsewhere...

tensorpipe/core/message.h
```diff
@@ -49,9 +51,7 @@ class Message final {
   std::vector<Payload> payloads;
 
   struct Tensor {
-    void* data{nullptr};
-    size_t length{0};
-
+    tensorpipe::Tensor data;
```
lw, line 54: I think that here we should keep both the newly-introduced `data` field but, for now, to preserve backwards compatibility, we could also keep the previous "raw" fields (`data` and `length`) and have the pipe convert these raw fields to a CpuTensor and then use that going forward. This would allow us to land this change, migrate PyTorch over to use the new field, and then drop the old ones.

Unfortunately this means that we need to find another name for the new `data` field, as otherwise it conflicts with the old one...

tensorpipe/core/pipe.cc
```diff
@@ -87,21 +87,25 @@ void parseDescriptorOfMessage(
...
+    tensorBeingAllocated.length = tensor.data.cpu.length;
     tensorBeingAllocated.channelName = pbTensorDescriptor.channel_name();
     // FIXME If the protobuf wasn't const we could move the string out...
     tensorBeingAllocated.descriptor = pbTensorDescriptor.channel_descriptor();
     message.tensors.push_back(std::move(tensor));
     op.tensors.push_back(std::move(tensorBeingAllocated));
-  }
-}
+  } // namespace
+} // namespace tensorpipe
```
lw, line 105: Clang-format mishap?

tensorpipe/python/tensorpipe.cc
```diff
@@ -187,9 +187,9 @@ std::shared_ptr<IncomingMessage> prepareToAllocate(
   std::vector<std::shared_ptr<IncomingTensor>> pyTensors;
   pyTensors.reserve(tpMessage.tensors.size());
   for (const auto& tpTensor : tpMessage.tensors) {
-    TP_DCHECK(tpTensor.data == nullptr);
-    pyTensors.push_back(
-        std::make_shared<IncomingTensor>(tpTensor.length, tpTensor.metadata));
+    TP_DCHECK(tpTensor.data.cpu.data == nullptr);
```
lw, line 190: Maybe even add a check that it is indeed a CPU tensor?

tensorpipe/test/channel/channel_test.h
```diff
@@ -79,45 +80,42 @@ class ChannelTest : public ::testing::TestWithParam<ChannelTestHelper*> {
...
-        ptr,
-        length,
+        tensor,
         [descriptorPromise{std::move(descriptorPromise)}](
             const tensorpipe::Error& error, std::string descriptor) {
           descriptorPromise->set_value(
               std::make_tuple(error, std::move(descriptor)));
         },
-        [promise{std::move(promise)}](const tensorpipe::Error& error) {
+        [promise{std::move(promise)}, tensor](const tensorpipe::Error& error) {
```
lw, line 102: Why do we need to capture it?  If it needs to stay alive for the duration of the send, shouldn't we make the channels take it by copy rather than by const-ref? (same below)

lw (commented):
Tip: the change of moving endpoint/descriptor/... out of the Channel class into the namespace, could have been done as a separate commit before this one since it's quite orthogonal and (as both changes require many updates spread across several files) it would have made reviewing a bit easier.

tensorpipe/channel/channel.cc
```diff
@@ -0,0 +1,36 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <tensorpipe/channel/channel.h>
+#include <tensorpipe/common/tensor.h>
+
+namespace tensorpipe {
+namespace channel {
+template <>
```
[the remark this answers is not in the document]
beauby: This is template specialization: we're implementing the method for the case `TTensor = CpuTensor`.

tensorpipe/common/tensor.h
beauby: Yes, I kept the name as it was in `Message`, but I'd go for `ptr` as well.

tensorpipe/common/tensor.h
```diff
@@ -0,0 +1,53 @@
...
+
+#if TENSORPIPE_HAS_CUDA_IPC_CHANNEL
+struct CudaTensor {
+  void* data{nullptr};
+  size_t length{0};
+  cudaStream_t stream{cudaStreamDefault};
+};
+#endif // TENSORPIPE_HAS_CUDA_IPC_CHANNEL
+
+struct Tensor {
```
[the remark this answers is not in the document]
beauby: `std::variant` is C++17 and we've been limiting ourselves to C++14 so far. We could probably backport it, but we don't have that many unions laying around so it might not be worth.

tensorpipe/channel/channel.cc
beauby: As we agreed we don't need backwards compatibility for the `Channel` interface, this `const_cast` itself disappeared, but the point remains: what previously was a `const void*` parameter is now a `void*` member of a struct. We could make the corresponding member of `CpuTensor` a `const void*` and introduce a `MutableCpuTensor` (or the other way around), but I'm not sure the benefits outweigh the cost.
An other option would be to introduce getters  in `CpuTensor` (i.e. `void* getPtr()` and `const void* getPtr() const`). What do you think?

tensorpipe/common/tensor.h
beauby: I'd be in favor of keeping them copyable, as they're intended as being lightweight. Should we change our stance, the compiler will tell us about every disallowed copy happening.

tensorpipe/channel/channel.cc
beauby: As we agreed we don't need backwards compatibility for the `Channel` interface, this `const_cast` itself disappeared, but the point remains: what previously was a `const void*` parameter is now a `void*` member of a struct. We could make the corresponding member of `CpuTensor` a `const void*` and introduce a `MutableCpuTensor` (or the other way around), but I'm not sure the benefits outweigh the cost.
An other option would be to introduce getters  in `CpuTensor` (i.e. `void* getPtr()` and `const void* getPtr() const`). What do you think?

tensorpipe/python/tensorpipe.cc
beauby: For some reason, adding `TP_DCHECK(tpTensor.data.type == tensorpipe::DeviceType::kCpu);` causes:
```
lto1: internal compiler error: in add_symbol_to_partition_1, at lto/lto-partition.c:155
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://bugs.gentoo.org/> for instructions.
lto-wrapper: fatal error: /usr/bin/c++ returned 1 exit status
compilation terminated.
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

tensorpipe/common/tensor.h
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+#if TENSORPIPE_HAS_CUDA_IPC_CHANNEL // TODO: Generic TENSORPIPE_HAS_CUDA.
+#include <cuda_runtime.h>
+#endif // TENSORPIPE_HAS_CUDA_IPC_CHANNEL
+
+namespace tensorpipe {
+
+enum class DeviceType {
+  kCpu = 0,
```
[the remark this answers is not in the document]
lw: You marked this resolved but didn't do it.

tensorpipe/common/tensor.h
lw: Agreed.

tensorpipe/python/tensorpipe.cc
lw: Well that's weird... Do other compilers also fail? Did you report the bug? Adding such a check is not a blocker but it would be nice and I'm also curious about getting to the bottom of this...

tensorpipe/channel/channel.cc
lw: Using getters was indeed what I was thinking about... It may make the code a bit less readable though, not sure how that compares with the added static safety...

This is not a decision that directly affects the pipe API right? There, a Message is always passed by value (or moved, so by rvalue ref), hence it's never const. The const-vs-non-const choice only affects the channel's API. However, the Tensor's getters would then also be visible to the user of the pipe (because they are included in the Message) and they may look weird there. Is this reason enough to not introduce getters and give up the const-safety?

tensorpipe/common/tensor.h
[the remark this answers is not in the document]
lw: Right now we must support C++14 because PyTorch does but eventually I believe we'll drop it and move to C++17 and it would be nice to use the most idiomatic data structure once we do do. Having a polyfill for std::variant would allow us basically to do a drop-in replacement with limited user disruption. But since some disruption will still exist, I also see the argument for rolling our own type here.

One consideration is that I have never used std::variant and thus don't know how comfortable it is, both in terms of readability (for our code and our users' one) and in terms of static safety (getting compile errors when we forget to handle a case, and stuff like that).

Would it make sense that, before landing this diff, we start writing the PyTorch PR that migrates PyTorch to this new API, and see how it looks like? And, similarly, also stack on top of this another PR that makes the pipe able to handle CUDA channels and see if it's easy to do? Or, if this is too much work, find some other way to compare how std::variant and our own type would compare.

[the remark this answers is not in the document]
lw: Also, it is currently a bit "redundant" to have a member field called `tensor` inside the `Message::Tensor` class, which is of type `FooTensor`. I'm wondering if we could find a different naming. Maybe call the field `handle` and use `CpuHandle` and `CudaHandle`? Open to suggestions...

tensorpipe/common/tensor.h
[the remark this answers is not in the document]
beauby: Sorry, probably a rebase gone wrong. Fixed now.

PR #206, opened 2020-09-04
Fix libnop headers being private in CMake

lw:
And a missing const_cast


tensorpipe/channel/helpers.cc
```diff
@@ -17,7 +17,8 @@ namespace channel {
 Channel::TDescriptor saveDescriptor(const AbstractNopHolder& object) {
   const size_t len = object.getSize();
   Channel::TDescriptor out(len, '\0');
-  NopWriter writer(reinterpret_cast<uint8_t*>(out.data()), len);
+  NopWriter writer(
+      const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(out.data())), len);
```
beauby, line 21: You could also use `reinterpret_cast<uint8_t*>(&out[0])`.

beauby (approved):

PR #208, opened 2020-09-06
Fix CMake build of CudaIpc channel.

beauby:
This wasn't caught by CI as we're currently not running GPU tests on CI.


lw:
I guess we should enable CI coverage of CUDA channels shortly then? To get automated signal for stuff like this...

lw (approved):

PR #213, opened 2020-09-09
Get rid of `channelName()` in channel test helper.

beauby:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #216 Add support for CUDA channels in the Pipe.
* #214 Refactor channel tests.
* **#213 Get rid of `channelName()` in channel test helper.**
* #212 Make Channel API accept buffer structs rather than raw pointers.
* #211 Add TENSORPIPE_SUPPORTS_CUDA flag.
* #210 Move channel types from `channel::Channel::` to `channel::`.

CUDA channel tests are currently disabled and this workaround will
not be needed anymore with the upcoming test refactor.

Differential Revision: [D23598038](https://our.internmc.facebook.com/intern/diff/D23598038)

PR #212, opened 2020-09-09
Make Channel API accept buffer structs rather than raw pointers.

beauby:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #216 Add support for CUDA channels in the Pipe.
* #214 Refactor channel tests.
* #213 Get rid of `channelName()` in channel test helper.
* **#212 Make Channel API accept buffer structs rather than raw pointers.**
* #211 Add TENSORPIPE_SUPPORTS_CUDA flag.
* #210 Move channel types from `channel::Channel::` to `channel::`.

+ Introduce buffer.h defining the buffer struct(s). The `CpuBuffer`
struct is always defined, while the `CudaBuffer` struct is defined
only when `TENSORPIPE_SUPPORTS_CUDA` is true.
+ Update all channels to take a `CpuBuffer` or `CudaBuffer` for
`send`/`recv` rather than a raw pointer and a length.
+ Make the base `Channel`/`Context` classes templated on `TBuffer`,
effectively creating two channel hierarchies (one for CPU channels,
one for CUDA channels).
+ Update the Pipe and the generic channel tests to use the new API. So
far, generic channel tests are CPU only, and tests for the CUDA IPC
channel are (temporarily) disabled. A subsequent PR will take care of
refactoring tests so that generic tests work for CUDA channels. An
other PR will add support for CUDA tensors in the Pipe.

Differential Revision: [D23598033](https://our.internmc.facebook.com/intern/diff/D23598033)

PR #214, opened 2020-09-09
Refactor channel tests.

beauby:
* #216 Add support for CUDA channels in the Pipe.
* **#214 Refactor channel tests.**
* #213 Get rid of `channelName()` in channel test helper.
* #212 Make Channel API accept buffer structs rather than raw pointers.
* #211 Add TENSORPIPE_SUPPORTS_CUDA flag.
* #210 Move channel types from `channel::Channel::` to `channel::`.

Make generic channel tests work with either CPU or CUDA channels:
+ Move from defining tests within the `TEST_P` macro to defining
standalone test classes. This is needed because `TEST_P` does not play
well with templated fixture classes. While it has the downside of
requiring an extra step to register a test after it has been
defined (by calling `CHANNEL_TEST_GENERIC(Foo);`), it improves
readability of tests by having them override `server()`/`client()`
methods rather than call `testConnection()` with two lambdas.
+ Data wrapping (for channel-type agnostic tests) is now done through
a templated `DataWrapper` class rather than through a virtual method
on the `ChannelTestHelper`.

Differential Revision: [D23598040](https://our.internmc.facebook.com/intern/diff/D23598040)


tensorpipe/test/channel/basic/basic_test.cc
```diff
@@ -11,7 +11,7 @@
 
 namespace {
 
-class BasicChannelTestHelper : public ChannelTestHelper {
+class BasicChannelTestHelper : public ChannelTestHelper<tensorpipe::CpuTensor> {
```
lw, line 14: Wondering if we should introduce some `CpuChannelTestHelper` shortcuts?

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
+  return future;
+}
+
+template <typename TTensor>
+class ChannelTest {
  public:
-  ChannelTest() : helper_(GetParam()), peers_(helper_->makePeerGroup()) {}
+  virtual void run(ChannelTestHelper<TTensor>* helper) {
+    helper_ = helper;
+    peers_ = helper_->makePeerGroup();
```
lw, line 171: Doing this in the run method means that any test case that does _not_ want to use the server/client method patter must re-do this, which doesn't scale well.

Also, the fact that the default run method "forces" the server/client pattern may be odd, as users then have to override it and make it do an entirely different thing.

What we could do instead is:
- give ChannelTest an "entry point" method, which is the one called from the `TEST_P` macros below.
- give ChannelTest a pure virtual "run" method, which is called by the above entry-point method.
- have a ClientServerChannelTest subclass, which implements that run method, and in them calls the server/client methods, which are defined as purely virtual.

This way someone who wants to do a non-server/client test can subclass ChannelTest, avoid having to set up the helper_ and peers_ fields, and then implement run (without overriding an already-existing implementation).

Note that if the above suggestion (of using the constructor) works then the entry point method wouldn't be doing anything, and we could simplify it and have TEST_P directly call run.

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
-      });
+  void server(std::shared_ptr<transport::Connection> conn) override {
+    std::shared_ptr<Context<TTensor>> ctx =
+        this->helper_->makeContext("server");
+    auto channel = ctx->createChannel(std::move(conn), Endpoint::kListen);
+
+    // Initialize with sequential values.
+    std::vector<uint8_t> data(dataSize);
+    std::iota(data.begin(), data.end(), 0);
+    DataWrapper<TTensor> wrappedData(data);
```
lw, line 45: What does creating a `DataWrapper` directly buy us, instead of using the helper's `makeBuffer` method which you removed?

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+    // Perform send and wait for completion.
+    std::future<std::tuple<Error, TDescriptor>> descriptorFuture;
+    std::future<Error> sendFuture;
+    std::tie(descriptorFuture, sendFuture) =
+        sendWithFuture(channel, wrappedData.tensor());
+    Error descriptorError;
+    TDescriptor descriptor;
+    std::tie(descriptorError, descriptor) = descriptorFuture.get();
+    EXPECT_FALSE(descriptorError) << descriptorError.what();
+    this->peers_->send(PeerGroup::kClient, descriptor);
```
lw, line 56: This sounds like a C++ thing I'm unaware of, but why do we suddenly need `this->`?

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+    this->peers_->join(PeerGroup::kClient);
+
+    ctx->join();
+  }
 };
 
-TEST_P(ChannelTest, ServerToClient) {
+CHANNEL_TEST_GENERIC(ClientToServer);
+
+// TEST_P(ChannelTest, ServerToClient) {
```
lw, line 95: What's this?

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+
+    ctx->join();
+  }
+};
+
+CHANNEL_TEST_GENERIC(ServerToClient);
+
+template <typename TTensor>
+class SendMultipleTensorsTest : public ChannelTest<TTensor> {
+  const int dataSize = 256 * 1024; // 256KB
```
lw, line 161: This was a `constexpr`, why the change?

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+
+    ctx->join();
+  }
+
+  void client(std::shared_ptr<transport::Connection> conn) override {
+    std::shared_ptr<Context<TTensor>> ctx =
+        this->helper_->makeContext("client");
+    auto channel = ctx->createChannel(std::move(conn), Endpoint::kConnect);
+
+    std::vector<std::unique_ptr<DataWrapper<TTensor>>> wrappedDataVec;
```
lw, line 207: Do we need to wrap the DataWrappers in unique_ptrs because they are not movable? Would it make sense perhaps to make them movable? As it stands, we're making plenty of heap allocations just so they can hold a few pointers...

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+
+    // Perform send and wait for completion.
+    std::future<std::tuple<Error, TDescriptor>> descriptorFuture;
+    std::future<Error> sendFuture;
+    std::tie(descriptorFuture, sendFuture) = sendWithFuture(
+        channel,
+        TTensor{
+            .ptr = nullptr,
+            .length = 0,
+        });
```
lw, line 375: This is not generic. We should either find a way to make it generic or, as there is no guarantee (in principle) that for every possible device type its data can be represented as a pointer, we shouldn't make this test generic but have an instance of Cpu and (if we so choose) one for Cuda.

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+  void server(std::shared_ptr<transport::Connection> conn) override {
+    std::shared_ptr<Context<TTensor>> ctx =
+        this->helper_->makeContext("server");
+    auto channel = ctx->createChannel(std::move(conn), Endpoint::kListen);
+
+    // Allocate a non-empty vector so that its .data() pointer is non-null.
+    std::vector<uint8_t> data(1);
+    DataWrapper<TTensor> wrappedData(data);
+    TTensor tensor = wrappedData.tensor();
+    tensor.length = 0;
```
lw, line 426: Similarly non-generic.

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+    // Perform send and wait for completion.
+    std::promise<std::tuple<Error, TDescriptor>> descriptorPromise;
+    std::promise<Error> sendPromise;
+    std::mutex mutex;
+    std::unique_lock<std::mutex> callerLock(mutex);
+    channel->send(
+        CpuTensor{
+            .ptr = data.data(),
+            .length = data.size(),
+        },
```
lw, line 518: Goes back to a comment I left on an earlier diff: why use "keyword" args instead of "positional" args? It seems we're mixing the two styles.

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,513 +15,568 @@
...
+        [&recvPromise, &mutex](const Error& error) {
+          std::unique_lock<std::mutex> calleeLock(mutex);
+          recvPromise.set_value(error);
+        });
+    callerLock.unlock();
+    Error recvError = recvPromise.get_future().get();
+    EXPECT_FALSE(recvError) << recvError.what();
+
+    // Validate contents of vector.
+    for (auto i = 0; i < dataSize; i++) {
```
lw, line 571: Also a reference to a previous comment: can we stick to dataSize or data.size() consistently?

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
```diff
@@ -17,70 +17,23 @@
 
 namespace {
 
-class CudaWrapper : public DataWrapper {
+class CudaChannelTestHelper : public ChannelTestHelper<tensorpipe::CudaTensor> {
```
lw, line 20: Shouldn't this be called `CudaIpcChannelTestHelper`? (Note the `Ipc`)

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
```diff
@@ -17,70 +17,23 @@
...
-
-//   std::shared_ptr<DataWrapper> makeBuffer(size_t len) override {
-//     return std::make_shared<CudaWrapper>(len);
-//   }
-// };
-
-// CudaChannelTestHelper helper;
-
-class CudaIpcChannelTest : public ChannelTest {};
+class CudaIpcChannelTest : public CudaChannelTest {};
```
lw, line 36: What is this doing exactly? I don't see anything equivalent in the other channels.

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
```diff
@@ -92,113 +45,115 @@ using namespace tensorpipe::channel;
...
+    TP_CUDA_CHECK(cudaFree(ptr));
+
+    this->peers_->done(PeerGroup::kClient);
+    this->peers_->join(PeerGroup::kClient);
+
+    ctx->join();
+  }
+};
+
+CHANNEL_TEST(CudaIpc, ReceiverWaitsForStartEvent);
```
lw, line 153: I thought the first argument of `CHANNEL_TEST` is the type of the tensor, which should be `Cuda` here. In fact, this test (`ReceiverWaitsForStartEvent`) should probably go in a CUDA test suite, and then this CUDA test suite should be instantiated once for CudaIpc. Right?

lw (commented):
I feel like this diff is doing two things at once: create two hierarchies of ChannelTest/ChannelTestHelper by templating them, and moving from the testConnection method to the server/client virtual methods (plus a bunch of other small things). It would have been clearer if they had been done in separate diffs.

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
beauby: The first argument of `CHANNEL_TEST` is a prefix for a test suite, for instance `Foo` -> `FooChannelTest`. It coincides with the type of the tensor in the case of generic test suites, but that's just a convenience.

tensorpipe/test/channel/channel_test.cc
beauby: With the current setting, the `makeBuffer()` method would simply forward its parameter to the constructor of `DataWrapper` and return that, it seems like an unnecessary indirection.

tensorpipe/test/channel/channel_test.cc
beauby: Removing it.

tensorpipe/test/channel/channel_test.cc
beauby: Yes, making them movable is probably the right thing to do here. Will update.

tensorpipe/test/channel/channel_test.cc
beauby: Sounds good.

tensorpipe/test/channel/channel_test.cc
beauby: Will fix.

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
-  std::shared_ptr<PeerGroup> peers_;
-
+template <typename TTensor>
+[[nodiscard]] std::pair<
+    std::future<
+        std::tuple<tensorpipe::Error, tensorpipe::channel::TDescriptor>>,
+    std::future<tensorpipe::Error>>
+sendWithFuture(
+    std::shared_ptr<tensorpipe::channel::Channel<TTensor>> channel,
+    TTensor tensor) {
```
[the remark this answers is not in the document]
beauby: As we moved to pass by copy for tensors, let's just keep it non const.

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
+        promise->set_value(error);
+      });
+  return future;
+}
+
+template <typename TTensor>
+class ChannelTest {
  public:
-  ChannelTest() : helper_(GetParam()), peers_(helper_->makePeerGroup()) {}
+  virtual void run(ChannelTestHelper<TTensor>* helper) {
```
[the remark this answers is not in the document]
beauby: Inheriting from `TestWithParams` here won't work, but I have an idea to avoid having to set those in `run()`. Will update the PR.

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
beauby: Good catch, will update.

tensorpipe/test/channel/basic/basic_test.cc
beauby: Will do.

tensorpipe/test/channel/channel_test.cc
lw: Damned C++, this is ADL all over again. I'm thankful I've learned about this the gentle way though...

tensorpipe/test/channel/channel_test.cc
lw: Ok, if it doesn't work don't bother (but maybe add a comment so that if anyone who comes around knows they can help).

tensorpipe/test/channel/channel_test.cc
lw: It looks like one comment I though I had made before has disappeared, I'm rewriting it now.

tensorpipe/test/channel/channel_test.cc
```diff
@@ -15,495 +15,556 @@
...
+  void server(std::shared_ptr<transport::Connection> conn) override {
+    std::shared_ptr<Context<TBuffer>> ctx =
+        this->helper_->makeContext("server");
+    auto channel = ctx->createChannel(std::move(conn), Endpoint::kListen);
+
+    // Perform send and wait for completion.
+    std::future<std::tuple<Error, TDescriptor>> descriptorFuture;
+    std::future<Error> sendFuture;
+    std::tie(descriptorFuture, sendFuture) =
+        sendWithFuture(channel, TBuffer{nullptr, 0});
```
lw, line 370: This is not a "generic" test, because it assumes that TBuffer can be constructed using two arguments: a pointer and a length. This may be the case with Cpu and CudaBuffer, but it's not guaranteed to always be the case. This test aims to test whether a channel can handle a null pointer but, in principle, a TBuffer may not even hold a pointer. So maybe it's better to make this become a Cpu-specific test? (And possibly create a copy of it for Cuda channels, if we decide we want Cuda channels to also be able to handle null pointers)

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
lw: Right, but (also relating to the comment above), this test isn't specific to the `cuda_ipc` channel, it could be run on all Cuda channels (and should work for all of them). So shouldn't we add it to the "generic" Cuda test suite, rather than to the one specific to this one channel?

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
+      std::move(descriptor),
+      std::move(tensor),
+      [promise{std::move(promise)}](const tensorpipe::Error& error) {
+        promise->set_value(error);
+      });
+  return future;
+}
+
+template <typename TTensor>
+class ChannelTest {
```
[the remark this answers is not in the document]
beauby: Good idea. I separated the base `ChannelTestCase<>` from its subclass `ClientServerChannelTestCase<>`, and renamed the test suites `CpuChannelTestSuite/CudaChannelTestSuite`.

tensorpipe/test/channel/channel_test.h
beauby: The problem with using a constructor (with a parameter) is that it won't be inherited. The current trade off I came up with is as follows: the helper is passed as a parameter to `run()` in every case, and when, for a specific pattern, the helper needs to be made available from other methods (which currently only happens in the ClientServerChannelTestCase), this is done in `run()`.

tensorpipe/test/channel/channel_test_cpu.cc
```diff
@@ -0,0 +1,200 @@
...
+CHANNEL_TEST(CpuChannelTestSuite, EmptyTensor);
+
+// This test wants to make sure that the "heavy lifting" of copying data isn't
+// performed inline inside the recv method as that would make the user-facing
+// read method of the pipe blocking.
+// However, since we can't really check that behavior, we'll check a highly
+// correlated one: that the recv callback isn't called inline from within the
+// recv method. We do so by having that behavior cause a deadlock.
+class CallbacksAreDeferredTest
+    : public ClientServerChannelTestCase<tensorpipe::CpuBuffer> {
```
lw, line 121: The tests above do not need the namespace qualifier
```suggestion
    : public ClientServerChannelTestCase<CpuBuffer> {
```

tensorpipe/test/channel/channel_test_cuda.cc
```diff
@@ -0,0 +1,121 @@
...
+#include <cuda_runtime.h>
+
+#include <tensorpipe/test/channel/channel_test.h>
+#include <tensorpipe/test/channel/kernel.cuh>
+
+using namespace tensorpipe;
+using namespace tensorpipe::channel;
+
+class ReceiverWaitsForStartEventTest
+    : public ClientServerChannelTestCase<tensorpipe::CudaBuffer> {
```
lw, line 18: Also probably not needed?
```suggestion
    : public ClientServerChannelTestCase<CudaBuffer> {
```

tensorpipe/test/channel/channel_test.h
```diff
@@ -73,6 +70,14 @@ class DataWrapper<tensorpipe::CudaBuffer> {
     }
   }
 
+  explicit DataWrapper(DataWrapper&& other)
```
lw, line 73: Nit: doesn't the "rule of 5" say that whenever one defines a move constructor one should also define a move assignment operator?

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
```diff
@@ -31,129 +30,17 @@ class CudaChannelTestHelper : public ChannelTestHelper<tensorpipe::CudaBuffer> {
   }
 };
 
-CudaChannelTestHelper helper;
+CudaIpcChannelTestHelper helper;
 
-class CudaIpcChannelTest : public CudaChannelTest {};
+class CudaIpcChannelTestSuite : public CudaChannelTestSuite {};
```
lw, line 35: Now there are no IPC-specific CUDA tests anymore. Do we still need this?

tensorpipe/test/channel/cuda_ipc/cuda_ipc_test.cc
```diff
@@ -31,129 +30,17 @@ class CudaChannelTestHelper : public ChannelTestHelper<tensorpipe::CudaBuffer> {
...
-INSTANTIATE_TEST_CASE_P(CudaIpc, CudaChannelTest, ::testing::Values(&helper));
 INSTANTIATE_TEST_CASE_P(
     CudaIpc,
-    CudaIpcChannelTest,
+    CudaChannelTestSuite,
+    ::testing::Values(&helper));
+INSTANTIATE_TEST_CASE_P(
+    CudaIpc,
+    CudaIpcChannelTestSuite,
     ::testing::Values(&helper));
```
lw: Same: do we still need this?

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
+class DataWrapper<tensorpipe::CpuTensor> {
  public:
-  virtual ~DataWrapper() = default;
-  virtual void* data() = 0;
-  virtual size_t size() = 0;
-  virtual void wrap(const void* ptr) = 0;
-  virtual void unwrap(void* ptr) = 0;
+  explicit DataWrapper(size_t length) : vector_(length) {}
+
+  explicit DataWrapper(const std::vector<uint8_t>& v) : vector_(v) {}
```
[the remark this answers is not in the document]
lw: You marked thus as resolved but you didn't change anything...

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
+
+  tensorpipe::CpuTensor tensor() const {
+    return tensorpipe::CpuTensor{
+        .ptr = const_cast<uint8_t*>(vector_.data()),
+        .length = vector_.size(),
+    };
+  }
+
+  std::vector<uint8_t> unwrap() {
```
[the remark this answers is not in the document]
lw: You marked thus as resolved but you didn't change anything...

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,69 +20,156 @@
...
+
+    return v;
   }
 
-  void unwrap(void* ptr) override {
-    ASSERT_EQ(data_, ptr);
+  ~DataWrapper() {
+    if (length_ > 0) {
+      EXPECT_EQ(cudaSuccess, cudaFree(cudaPtr_));
+    }
```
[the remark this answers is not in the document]
lw: You marked thus as resolved but you didn't change anything...

tensorpipe/test/channel/channel_test.h
lw: Isn't there a way to have subclasses inherit constructors? Something like adding `using SuperClass;` in their body. It's not automatic (a subclass may "forget" to do so) but at least it avoids duplicating the initialization code...

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,71 +20,168 @@
...
-  virtual void wrap(const void* ptr) = 0;
-  virtual void unwrap(void* ptr) = 0;
+  explicit DataWrapper(size_t length) : vector_(length) {}
+
+  explicit DataWrapper(const std::vector<uint8_t>& v) : vector_(v) {}
+
+  tensorpipe::CpuBuffer buffer() const {
+    return tensorpipe::CpuBuffer{
+        .ptr = const_cast<uint8_t*>(vector_.data()),
+        .length = vector_.size(),
```
lw, line 36: Didn't you say that this is a C++-20 thing?

tensorpipe/test/channel/channel_test.h
```diff
@@ -20,71 +20,168 @@
...
 
-  void wrap(const void* ptr) override {
-    data_ = const_cast<void*>(ptr);
+  explicit DataWrapper(DataWrapper&& other)
+      : cudaPtr_(other.cudaPtr_),
+        length_(other.length_),
+        stream_(other.stream_) {
+    other.cudaPtr_ = nullptr;
+    other.length_ = 0;
   }
```
lw, line 79: Once you destroy the stream in the destructor you may want to reset it to `cudaStreamDefault` here.

tensorpipe/test/channel/channel_test.h
beauby: This is my understanding of https://en.cppreference.com/w/cpp/language/aggregate_initialization. Basically it's a C99 thing that hasn't made it in the C++ standard before C++20, but most compilers accept it IIUC. Anyways, reverting to something officially supported.

tensorpipe/test/channel/channel_test.h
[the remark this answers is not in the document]
beauby: My bad, some rebasing gone wrong.

tensorpipe/test/channel/channel_test.h
beauby: I considered the `using SuperClass::SuperClass` route but it seemed lesss convenient than the current solution (where subclasses only define test-related methods, being `run()` in the general case, or `server()/client()` when using the `ClientServer` pattern. Note that exposing the `helper` parameter of `run()` as a class member is not needed in all cases (currently only with ClientServer), and has more to do with how `ClientServerChannelTestCase` works than general "intialization of a test case".

tensorpipe/test/channel/channel_test.h
```diff
@@ -76,14 +74,22 @@ class DataWrapper<tensorpipe::CudaBuffer> {
...
+    other.stream_ = cudaStreamDefault;
+  }
+
+  DataWrapper& operator=(DataWrapper&& other) {
+    cudaPtr_ = other.cudaPtr_;
+    length_ = other.length_;
+    stream_ = other.stream_;
+    other.cudaPtr_ = nullptr;
+    other.length_ = 0;
+    other.stream_ = cudaStreamDefault;
```
lw, line 86: Sorry to keep nit-picking, but this isn't a valid move assignment operator: if `this` was holding a resource, this operator is "forgetting" that resource without freeing it properly. One "trick" for a valid move constructor is to `std::swap(...)` all fields. This way, the object that we're moving from will end up destroying the resource. (Note that a similar trick also works for the move constructor, although it's not necessary there because `this` can't possibly be holding a resource yet)

tensorpipe/test/channel/channel_test.h
```diff
@@ -76,14 +74,22 @@ class DataWrapper<tensorpipe::CudaBuffer> {
...
   tensorpipe::CudaBuffer buffer() const {
-    return tensorpipe::CudaBuffer{
-        .ptr = cudaPtr_,
-        .length = length_,
-        .stream = stream_,
-    };
+    return tensorpipe::CudaBuffer{cudaPtr_, length_, stream_};
   }
 
   std::vector<uint8_t> unwrap() {
```
lw, line 95: Return by const-ref here too?

lw (approved):
Only a couple of minor things left, otherwise very good!

tensorpipe/test/channel/channel_test.h
beauby: Returning a const ref here is UB, as the vector is stack-allocated. Moreover, copy-elision should prevent actually copying the object.

lw: Sorry, you're right.

lw (commented):
Perfect!

lw (approved):

PR #216, opened 2020-09-09
Add support for CUDA channels in the Pipe.

beauby:
* **#216 Add support for CUDA channels in the Pipe.**
* #214 Refactor channel tests.
* #213 Get rid of `channelName()` in channel test helper.
* #212 Make Channel API accept buffer structs rather than raw pointers.
* #211 Add TENSORPIPE_SUPPORTS_CUDA flag.
* #210 Move channel types from `channel::Channel::` to `channel::`.

The approach we are taking is basically duplicating everything that is
done for CPU channels. There is (quite some) room for refactoring, but
we can always circle back later as the only thing that has to remain
stable is the public API.

Differential Revision: [D23598035](https://our.internmc.facebook.com/intern/diff/D23598035)


tensorpipe/core/pipe.cc
```diff
@@ -53,6 +53,7 @@ struct ReadOperation {
   };
   std::vector<Payload> payloads;
   struct Tensor {
+    DeviceType deviceType;
```
lw, line 56: Why not make two `Tensor` structs and store them in a `tensors` and a `cudaTensors` fields?

tensorpipe/core/pipe.cc
```diff
@@ -85,21 +86,45 @@ void parseDescriptorOfMessage(ReadOperation& op, const Packet& nopPacketIn) {
...
-    tensorBeingAllocated.length = tensor.tensor.cpu.length;
+    tensorBeingAllocated.length = nopTensorDescriptor.sizeInBytes;
+    tensorBeingAllocated.deviceType = nopTensorDescriptor.deviceType;
     tensorBeingAllocated.channelName = nopTensorDescriptor.channelName;
     // FIXME If the nop object wasn't const we could move the string out...
     tensorBeingAllocated.descriptor = nopTensorDescriptor.channelDescriptor;
-    message.tensors.push_back(std::move(tensor));
     op.tensors.push_back(std::move(tensorBeingAllocated));
+
+    switch (nopTensorDescriptor.deviceType) {
```
lw, line 97: Why not also store cpu and cuda tensors on separate fields in the nop types?

tensorpipe/core/pipe.cc
```diff
@@ -85,21 +86,45 @@ void parseDescriptorOfMessage(ReadOperation& op, const Packet& nopPacketIn) {
...
+      case DeviceType::kCpu: {
+        Message::Tensor tensor{
```
lw, line 99: We could have the cases of the switch just create a tensorpipe::Tensor and then pack it into a tensorpipe::Message::Tensor outside of the switch. This way we wouldn't have to duplicate the metadata stuff, for example.

tensorpipe/core/pipe.cc
```diff
@@ -1264,13 +1405,54 @@ void Pipe::Impl::onReadWhileServerWaitingForBrochure_(
           impl.onAcceptWhileServerWaitingForChannel_(
               channelName, std::move(transport), std::move(connection));
         }));
-    channelRegistrationIds_[channelName] = token;
+    channelRegistrationIds_["cpu_" + channelName] = token;
```
lw, line 1408: Could we just use two separate fields for cpu and cuda? It is quite cleaner than tampering with strings, and also statically safer.

lw (commented):
Didn't finish this review as I'd like to propose a different approach: instead of using enums everywhere and switching on them and catching bad values at runtime, can we just duplicate all the fields that we used for cpu and add a copy of them for cuda? This way there's a compile-time difference between them, and a stronger guarantee.

As for the overhead coming from the duplication, I think we may be able to address it with some macros. I also don't like the tons of `#if TENSORPIPE_HAS_CUDA` checks we have now, so perhaps (later on) we could introduce something like what PyTorch does to execute some code over all the dtypes: https://github.com/pytorch/pytorch/blob/3aec1185e0643b2249984577e5f8a822b2adccde/c10/core/ScalarType.h#L112

PR #228, opened 2020-10-08
Fix device-specific fields in the pipe.

beauby:


tensorpipe/core/buffer_helpers.h
```diff
@@ -11,30 +11,24 @@
...
-    TP_BUFFER_CUDA_FIELD_AND_ACCESSOR(cuda_type);         \
+#define TP_DEVICE_FIELD(cpu_type, cuda_type)                               \
+  struct {                                                                 \
+    template <typename TBuffer>                                            \
+    constexpr auto& get() {                                                \
+      if (std::is_same<TBuffer, CpuBuffer>::value) {                       \
+        return cpu;                                                        \
+      }                                                                    \
+      TP_IF_CUDA(                                                          \
+          if (std::is_same<TBuffer, CudaBuffer>::value) { return cuda; }); \
```
lw, line 28: Not sure if the TP_IF_CUDA is buying us much in terms of readability in this particular instance, what do you think?

lw (approved):

tensorpipe/core/buffer_helpers.h
beauby: We need it here because `CudaBuffer` is not defined unless `TENSORPIPE_SUPPORTS_CUDA` is defined.

lw: My point was that doing the full `#if TENSORPIPE_SUPPORTS_CUDA` thing is perhaps clearer here than passing an `if (...)` expression as an argument of a function-like macro. For example, that confuses clang-format, which formats it differently than the previous if. I found it confusing, but if you like it I'm ok to stick to it.

beauby: That would not work because we'd need an `#if` inside the macro definition.

lw: Oh, right, I keep forgetting that this doesn't work. Fine to keep it like this then.

PR #232, opened 2020-10-19
Add CI for GPU.

beauby:
* #235 [WIP] Add cuda_xth channel.
* #234 [WIP] Add cuda_basic channel.
* #233 Add missing include in channel_test.h.
* **#232 Add CI for GPU.**

Differential Revision: [D24758822](https://our.internmc.facebook.com/intern/diff/D24758822)


.circleci/config.yml
```diff
@@ -61,6 +61,47 @@ jobs:
           command: |
             cd build
             make install
+  build_gpu:
+    machine:
+      resource_class: gpu.nvidia.small
+      image: ubuntu-1604-cuda-10.1:201909-23
```
lw, line 67: What images are available? I very much suspect we need to test with multiple versions of CUDA, at least the ones supported/tested by PyTorch.

.circleci/config.yml
```diff
@@ -61,6 +61,47 @@ jobs:
...
+            cd build
+            cmake ../ \
+              -DCMAKE_C_FLAGS="-Werror -Wno-deprecated-declarations" \
+              -DCMAKE_CXX_FLAGS="-Werror -Wno-deprecated-declarations" \
+              -DTP_ENABLE_SHM=OFF \
+              -DTP_ENABLE_CMA=OFF \
+              -DTP_USE_CUDA=ON \
+              -DTP_ENABLE_CUDA_IPC=ON \
+              -DTP_ENABLE_IBV=OFF \
+              -DTP_BUILD_TESTING=ON
```
lw, line 93: Are we disabling all the other ones because they are failing, or because you want this to be focused on CUDA only?

lw (approved):
I hadn't accepted because CI was broken but I see it's being fixed in the following PR.

.circleci/config.yml
beauby: There's only ubuntu 16.04 with either cuda 10.1 or 9.2 available. If we want to test against multiple versions of cuda, the easiest way will be to build docker containers. The scope of this PR is simply to get _some_ signal (i.e. ensure it builds fine and it is not broken in an obvious way), but we can discuss about more thorough coverage.

.circleci/config.yml
beauby: A bit of both actually, the other ones except for SHM were disabled on the cpu Linux builds, and shm was failing here.

PR #243, opened 2020-11-18
Set device before calling cudaStreamWaitEvent() in CUDA IPC channel.

beauby:
Summary: A call to cudaStreamWaitEvent() should happen on the device the stream is defined on, according to CUDA docs.

Differential Revision: D24859542

### DIFF
--- a/tensorpipe/test/channel/channel_test_cuda.cc
+++ b/tensorpipe/test/channel/channel_test_cuda.cc
@@ -10,6 +10,7 @@
 
 #include <tensorpipe/channel/cuda_context.h>
 #include <tensorpipe/test/channel/channel_test.h>
+#include <tensorpipe/test/test_environment.h>
 #include <tensorpipe/test/channel/kernel.cuh>
 
 using namespace tensorpipe;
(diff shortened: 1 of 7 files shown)


tensorpipe/test/test_environment.h
```diff
@@ -0,0 +1,18 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#if TENSORPIPE_SUPPORTS_CUDA
+#include <tensorpipe/common/cuda.h>
```
lw, line 12: I don't think this header is needed? Maybe move it to the .cc file?

tensorpipe/test/test_environment.cc
```diff
@@ -0,0 +1,45 @@
...
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <tensorpipe/common/cuda.h>
+#include <tensorpipe/common/defs.h>
+#include <unistd.h>
+#endif // TENSORPIPE_SUPPORTS_CUDA
+
+int TestEnvironment::numCudaDevices() {
+  static int count = -1;
+  if (count == -1) {
```
lw, line 24: Do we need to worry about race conditions when multiple threads call this function at the same time? (The C++ runtime guarantees atomicity on the assignment of a local static variable but not on subsequent modifications).

tensorpipe/test/test_environment.h
```diff
@@ -0,0 +1,18 @@
...
+#endif // TENSORPIPE_SUPPORTS_CUDA
+
+class TestEnvironment {
+ public:
+  static int numCudaDevices();
```
lw, line 17: I'm wondering if we should rather make this a member function (a method) rather than a static one. For now a static function makes sense, so that we don't have to pass an instance of the environment around everywhere, but perhaps one day once we add more fields to this we'll want to run a test multiple time with different environments. (In general I guess this relates to my tendency to avoid global state as much as possible).

TL;DR: this is perfectly fine for now, but there's food for thought.

tensorpipe/test/channel/channel_test_cuda.cc
```diff
@@ -120,3 +121,118 @@ class ReceiverWaitsForStartEventTest
...
 CHANNEL_TEST(CudaChannelTestSuite, ReceiverWaitsForStartEvent);
+
+class SendAcrossDevicesTest : public ClientServerChannelTestCase<CudaBuffer> {
+  static constexpr size_t kSize = 1024;
+
+ public:
+  void run(ChannelTestHelper<CudaBuffer>* helper) override {
+    if (TestEnvironment::numCudaDevices() < 2) {
+      GTEST_SKIP() << "Skipping test requiring >=2 CUDA devices.";
+      return;
```
lw, line 132: Do we need to return? One of the gtest examples seems to hint GTEST_SKIP will somehow skips what comes after (I guess it raises an exception). https://github.com/google/googletest/blob/master/googletest/test/gtest_skip_test.cc#L37-L40

lw (approved):
Thanks for adding a test (and the infrastructure needed to get it to work) and to make it harder to forget this in the future!

CircleCI "correctly" skips the test, but I'm assuming you ran it for both XTH and IPC manually to ensure it passes?

tensorpipe/test/test_environment.cc
beauby: The race condition here is pretty harmless, as `cudaGetDeviceCount()` will always return the same value, so worst case it will be called multiple times.

tensorpipe/test/test_environment.h
beauby: Good catch, removed.

tensorpipe/test/channel/channel_test_cuda.cc
beauby: Updated.

PR #244, opened 2020-11-20
Update googletest submodule commit

beauby:
Summary: The googletest version in fbcode was updated (D24515800), hence we can use the newer version in OSS.

Differential Revision: D25121752


lw (approved):
Awesome!

PR #245, opened 2020-11-22
Add cuda_basic channel. (#234)

beauby:
Summary: Pull Request resolved: https://github.com/pytorch/tensorpipe/pull/234

Test Plan: Imported from OSS

Differential Revision: D24758786

Pulled By: beauby

### DIFF
--- a/tensorpipe/channel/cuda_basic/channel.cc
+++ b/tensorpipe/channel/cuda_basic/channel.cc
@@ -1,0 +1,425 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <tensorpipe/channel/cuda_basic/channel.h>
+
+#include <algorithm>
+#include <condition_variable>
+#include <list>
+#include <mutex>
+
+#include <tensorpipe/channel/cuda_basic/context_impl.h>
+#include <tensorpipe/channel/error.h>
+#include <tensorpipe/channel/helpers.h>
+#include <tensorpipe/common/callback.h>
+#include <tensorpipe/common/cuda.h>
+#include <tensorpipe/common/defs.h>
+#include <tensorpipe/common/error.h>
+#include <tensorpipe/common/error_macros.h>
+#include <tensorpipe/common/system.h>
+#include <tensorpipe/transport/connection.h>
+
+namespace tensorpipe {
+namespace channel {
+namespace cuda_basic {
+
+namespace {
+
+std::shared_ptr<uint8_t> makeCudaPinnedBuffer(size_t length) {
+  void* ptr;
+  TP_CUDA_CHECK(cudaMallocHost(&ptr, length));
+  return std::shared_ptr<uint8_t>(
+      reinterpret_cast<uint8_t*>(ptr),
+      [](uint8_t* ptr) { TP_CUDA_CHECK(cudaFreeHost(ptr)); });
+}
+
+} // namespace
+
+class Channel::Impl : public std::enable_shared_from_this<Channel::Impl> {
+ public:
+  Impl(
+      std::shared_ptr<Context::PrivateIface>,
+      std::shared_ptr<CpuChannel>,
+      std::string);
+
+  // Called by the channel's constructor.
+  void init();
+
+  void send(
+      CudaBuffer buffer,
+      TDescriptorCallback descriptorCallback,
+      TSendCallback callback);
+
+  void recv(TDescriptor descriptor, CudaBuffer buffer, TRecvCallback callback);
+
+  // Tell the channel what its identifier is.
+  void setId(std::string id);
+
+  void close();
+
+ private:
+  OnDemandDeferredExecutor loop_;
+
+  void initFromLoop();
+
+  // Send memory region to peer.
+  void sendFromLoop(
+      CudaBuffer buffer,
+      TDescriptorCallback descriptorCallback,
+      TSendCallback callback);
+
+  // Receive memory region from peer.
+  void recvFromLoop(
+      TDescriptor descriptor,
+      CudaBuffer buffer,
+      TRecvCallback callback);
+
+  void closeFromLoop();
+
+  void setError(Error error);
+
+  void setIdFromLoop(std::string id);
+
+  // Helper function to process transport error.
+  // Shared between read and write callback entry points.
+  void handleError();
+
+  // Proxy static method for cudaStreamAddCallback(), which does not accept
+  // lambdas.
+  static void CUDART_CB signalOperationReady_(cudaStream_t, cudaError_t, void*);
+  void processOperations();
+
+  std::thread opThread_;
+  std::mutex opMutex_;
+  std::condition_variable opCondVar_;
+
+  struct Operation {
+    std::function<void(cudaError_t)> process;
+    std::function<void(cudaError_t)> signalReady;
+    bool ready{false};
+    cudaError_t cudaError;
+  };
+
+  std::deque<Operation> sendOperations_;
+  std::deque<Operation> recvOperations_;
+
+  std::shared_ptr<Context::PrivateIface> context_;
+  std::shared_ptr<CpuChannel> cpuChannel_;
+  std::unordered_map<uint64_t, std::function<void(cudaError_t)>>
+      cudaSendCallbacks_;
+  std::unordered_map<uint64_t, std::function<void(cudaError_t)>>
+      cudaRecvCallbacks_;
+  Error error_{Error::kSuccess};
+
+  ClosingReceiver closingReceiver_;
+
+  // Increasing identifier for send operations.
+  uint64_t nextTensorBeingSent_{0};
+
+  // Increasing identifier for recv operations.
+  uint64_t nextTensorBeingReceived_{0};
+
+  // An identifier for the channel, composed of the identifier for the context,
+  // combined with an increasing sequence number. It will only be used for
+  // logging and debugging purposes.
+  std::string id_;
+
+  LazyCallbackWrapper<Impl> lazyCallbackWrapper_{*this, this->loop_};
+
+  // For some odd reason it seems we need to use a qualified name here...
+  template <typename T>
+  friend class tensorpipe::LazyCallbackWrapper;
+};
+
+Channel::Channel(
+    ConstructorToken /* unused */,
+    std::shared_ptr<Context::PrivateIface> context,
+    std::shared_ptr<CpuChannel> cpuChannel,
+    std::string id)
+    : impl_(std::make_shared<Impl>(
+          std::move(context),
+          std::move(cpuChannel),
+          std::move(id))) {
+  impl_->init();
+}
+
+Channel::Impl::Impl(
+    std::shared_ptr<Context::PrivateIface> context,
+    std::shared_ptr<CpuChannel> cpuChannel,
+    std::string id)
+    : context_(std::move(context)),
+      cpuChannel_(std::move(cpuChannel)),
+      closingReceiver_(context_, context_->getClosingEmitter()),
+      id_(std::move(id)) {}
+
+void Channel::Impl::init() {
+  loop_.deferToLoop([this]() { initFromLoop(); });
+}
+
+void Channel::Impl::initFromLoop() {
+  TP_DCHECK(loop_.inLoop());
+  closingReceiver_.activate(*this);
+  opThread_ = std::thread([this]() { processOperations(); });
+}
+
+void Channel::send(
+    CudaBuffer buffer,
+    TDescriptorCallback descriptorCallback,
+    TSendCallback callback) {
+  impl_->send(buffer, std::move(descriptorCallback), std::move(callback));
+}
+
+void Channel::Impl::send(
+    CudaBuffer buffer,
+    TDescriptorCallback descriptorCallback,
+    TSendCallback callback) {
+  loop_.deferToLoop([this,
+                     buffer,
+                     descriptorCallback{std::move(descriptorCallback)},
+                     callback{std::move(callback)}]() mutable {
+    sendFromLoop(buffer, std::move(descriptorCallback), std::move(callback));
+  });
+}
+
+void Channel::Impl::processOperations() {
+  setThreadName("TP_CUDA_BASIC_loop");
+  static int sendOpId = 0;
+  static int recvOpId = 0;
+
+  for (;;) {
+    std::unique_lock<std::mutex> lock(opMutex_);
+    opCondVar_.wait(lock);
+
+    while (!sendOperations_.empty()) {
+      auto& op = sendOperations_.front();
+      if (!op.ready) {
+        break;
+      }
+
+      op.process(op.cudaError);
+      sendOperations_.pop_front();
+    }
+
+    while (!recvOperations_.empty()) {
+      auto& op = recvOperations_.front();
+      if (!op.ready) {
+        break;
+      }
+      op.process(op.cudaError);
+      recvOperations_.pop_front();
+      recvOpId++;
+    }
+
+    if (error_ && sendOperations_.empty() && recvOperations_.empty()) {
+      break;
+    }
+  }
+}
+
+// Send memory region to peer.
+void Channel::Impl::sendFromLoop(
+    CudaBuffer buffer,
+    TDescriptorCallback descriptorCallback,
+    TSendCallback callback) {
+  TP_DCHECK(loop_.inLoop());
+
+  const uint64_t sequenceNumber = nextTensorBeingSent_++;
+  TP_VLOG(4) << "Channel " << id_ << " received a send request (#"
+             << sequenceNumber << ")";
+
+  if (error_) {
+    descriptorCallback(error_, std::string());
+    callback(error_);
+    return;
+  }
+
+  TP_VLOG(4) << "Channel " << id_
+             << " is copying buffer from CUDA device to CPU";
+  auto tmpBuffer = makeCudaPinnedBuffer(buffer.length);
+  TP_CUDA_CHECK(cudaMemcpyAsync(
+      tmpBuffer.get(),
+      buffer.ptr,
+      buffer.length,
+      cudaMemcpyDeviceToHost,
+      buffer.stream));
+
+  sendOperations_.emplace_back();
+  auto& op = sendOperations_.back();
+  op.process = [this,
+                sequenceNumber,
+                buffer,
+                tmpBuffer{std::move(tmpBuffer)},
+                descriptorCallback{std::move(descriptorCallback)},
+                callback{std::move(callback)}](cudaError_t cudaError) {
+    TP_VLOG(4) << "Channel " << id_
+               << " is done copying buffer from CUDA device to CPU";
+    TP_CUDA_CHECK(cudaError);
+    CpuBuffer cpuBuffer{tmpBuffer.get(), buffer.length};
+    cpuChannel_->send(
+        cpuBuffer,
+        std::move(descriptorCallback),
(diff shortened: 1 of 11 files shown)


tensorpipe/common/cuda_loop.h
```diff
@@ -0,0 +1,55 @@
...
+
+  struct Operation {
+    TCudaCallback callback;
+    cudaError_t error;
+  };
+
+ public:
+  CudaLoop();
+
+  ~CudaLoop();
```
lw, line 33: This may need a `close` and `join` methods too.

tensorpipe/common/cuda_loop.h
```diff
@@ -0,0 +1,55 @@
...
+  std::deque<Operation> operations_;
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  uint64_t pendingOperations_ = 0;
+  bool joined_ = false;
+
+  // Proxy static method for cudaStreamAddCallback(), which does not accept
+  // lambdas.
+  static void CUDART_CB runCudaCallback(
+      cudaStream_t /* unused */,
```
lw, line 50: Nit: the fact that a param is unused is an "implementation detail" of the definition. It might be nice to hide this detail from the declaration of the function.

tensorpipe/common/cuda_loop.cc
```diff
@@ -0,0 +1,78 @@
...
+
+#include <tensorpipe/common/cuda_loop.h>
+
+#include <tensorpipe/common/cuda.h>
+#include <tensorpipe/common/system.h>
+
+namespace tensorpipe {
+
+CudaLoop::CudaLoop() {
+  std::unique_lock<std::mutex> lock(mutex_);
```
lw, line 17: Nit: no point in acquiring the lock in the constructor: no one else has yet a pointer to this class, so no one can call any method on it.

tensorpipe/common/cuda_loop.cc
```diff
@@ -0,0 +1,78 @@
...
+  thread_ = std::thread([this, lock{std::move(lock)}]() mutable {
```
lw, line 18: As there's a lot of logic in this lambda, maybe make it a real method to help readability?

tensorpipe/common/cuda_loop.cc
```diff
@@ -0,0 +1,78 @@
...
+    std::unique_lock<std::mutex> lock(mutex_);
+    joined_ = true;
+    cv_.notify_all();
+  }
+  thread_.join();
+}
+
+void CudaLoop::addCallback(cudaStream_t stream, TCudaCallback callback) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  TP_THROW_ASSERT_IF(joined_);
```
lw, line 54: I wouldn't be surprised if someone actually tried to add a callback once this is joined (since each channel has its own OnDemandDeferredExecutor, here's what could happen: the context closes, and destroys this loop, and enqueues a close on all channels, but some channels have some other operations already queued up in their on-demand loops, which will run before they "are informed" that they must close. these ops might try to add a callback here, which would fail). Another approach could be to just call the callback inline with some error.

tensorpipe/common/cuda_loop.cc
```diff
@@ -0,0 +1,78 @@
...
+  ++pendingOperations_;
+  auto it = cudaCallbacks_.emplace(cudaCallbacks_.end());
+  *it =
+      [this, it, callback{std::move(callback)}](cudaError_t cudaError) mutable {
+        std::unique_lock<std::mutex> lock(mutex_);
+        TP_THROW_ASSERT_IF(joined_);
```
lw, line 60: Oh and this too can _definitely_ be triggered. Whenever the loop is closed while there are still pending callbacks, we will wait for CUDA to call these callbacks, in which case they would hit this.

tensorpipe/common/cuda_loop.h
```diff
@@ -0,0 +1,55 @@
...
+  CudaLoop();
+
+  ~CudaLoop();
+
+  // TODO: device parameter?
+  void addCallback(cudaStream_t stream, TCudaCallback callback);
+
+ private:
+  std::thread thread_;
+  std::list<TCudaCallback> cudaCallbacks_;
```
lw, line 40: This list is used to store the callbacks in between the moment the user passes them to us and the moment CUDA tells us they're ready, right? So, during that time, "the ball is in CUDA's court", it's responsible for calling back into us so we can clean this. If we _really_ trusted CUDA we could just store the callback in a unique_ptr, pass it as the callbackPtr, "release" it on our side (AKA leak it), and then re-acquire it when CUDA calls into us.

I'm not saying to do so, as I'm not yet trusting CUDA that much, it's just a thought.

tensorpipe/common/cuda_loop.cc
```diff
@@ -0,0 +1,78 @@
...
+CudaLoop::CudaLoop() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  thread_ = std::thread([this, lock{std::move(lock)}]() mutable {
+    setThreadName("TP_CUDA_callback_loop");
+    for (;;) {
+      if (operations_.empty()) {
+        if (joined_ && pendingOperations_ == 0) {
+          break;
+        } else {
+          cv_.wait(lock);
```
lw, line 25: Nit: some linters suggest to always pass a predicate to `wait()` so that the stdlib will save you from spurious waits or from going to sleep when you could be doing something. In this case the condition is actually somewhat complicated, so perhaps doing it that way would be less readable. You decide.

tensorpipe/common/cuda_loop.cc
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      std::deque<Operation> operations;
+      std::swap(operations, operations_);
+
+      lock.unlock();
+      for (auto& op : operations) {
+        op.callback(op.error);
+      }
+      lock.lock();
```
lw, line 36: Nit: there's probably no strict need to release the lock here, since I guess these callbacks will be produced through `eagerCallbackWrapper`, which defers to the loop, so there's no risk of the callback calling back into this class. But releasing the lock is probably better as it reduces contention, so this was just an FYI.

tensorpipe/channel/cuda_basic/context.cc
```diff
@@ -0,0 +1,148 @@
...
+      // TODO: Lazy initialization of cuda loop.
+      cudaLoop_(std::make_shared<CudaLoop>()) {}
+
+void Context::close() {
+  impl_->close();
+}
+
+void Context::Impl::close() {
+  if (!closed_.exchange(true)) {
+    closingEmitter_.close();
```
lw, line 83: I think we need to close the cpuContext here, don't we?

lw: And the CudaLoop too (once it has a close method)

tensorpipe/channel/cuda_basic/context.cc
```diff
@@ -0,0 +1,148 @@
...
+void Context::join() {
+  impl_->join();
+}
+
+void Context::Impl::join() {
+  close();
+
+  if (!joined_.exchange(true)) {
+    cpuContext_->join();
+  }
```
lw, line 96: And here we should join the CudaLoop I guess

tensorpipe/channel/cuda_basic/context.cc
```diff
@@ -0,0 +1,148 @@
...
+  // their identifiers based off this context's identifier. They will only be
+  // used for logging and debugging.
+  std::atomic<uint64_t> channelCounter_{0};
+};
+
+Context::Context(std::shared_ptr<CpuContext> cpuContext)
+    : impl_(std::make_shared<Impl>(std::move(cpuContext))) {}
+
+Context::Impl::Impl(std::shared_ptr<CpuContext> cpuContext)
+    : domainDescriptor_("any"),
```
lw, line 72: Shouldn't this be the domain descriptor of the cpuChannel? In fact, we might not even have to store it, we can just forward any call to domainDescriptor() to the cpuChannel.

tensorpipe/channel/cuda_basic/context.cc
```diff
@@ -0,0 +1,148 @@
...
+  ~Impl() override = default;
+
+ private:
+  std::string domainDescriptor_;
+  std::atomic<bool> closed_{false};
+  std::atomic<bool> joined_{false};
+  ClosingEmitter closingEmitter_;
+
+  std::shared_ptr<CpuContext> cpuContext_;
+  std::shared_ptr<CudaLoop> cudaLoop_;
```
lw, line 56: Is there a reason to wrap this in a shared_ptr?

lw: I guess it's so that channels can keep a copy of the shared_ptr and keep the CudaLoop alive? In principle there shouldn't be a need for it: channels already hold a shared_ptr to the Context::Impl and thus keep that alive.

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -0,0 +1,358 @@
...
+#include <tensorpipe/common/error_macros.h>
+#include <tensorpipe/transport/connection.h>
+
+namespace tensorpipe {
+namespace channel {
+namespace cuda_basic {
+
+namespace {
+
+std::shared_ptr<uint8_t> makeCudaPinnedBuffer(size_t length) {
```
lw, line 33: Nit: any reason for this to be a shared_ptr rather than a unique_ptr? I know that declaring a custom deleter through a lambda is easier than defining a custom struct, but I like the stronger guarantees of unique_ptr. Also, this could go in common/cuda.h since it seems quite reusable.

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -0,0 +1,358 @@
...
+          CpuBuffer cpuBuffer{tmpBuffer.get(), buffer.length};
+          // Keep tmpBuffer alive until cpuChannel_ is done sending it over.
+          callback = eagerCallbackWrapper_(
+              [tmpBuffer{std::move(tmpBuffer)}, callback{std::move(callback)}](
+                  Impl& impl) { callback(impl.error_); });
+          cpuChannel_->send(
+              cpuBuffer, std::move(descriptorCallback), std::move(callback));
+        });
+      };
+  cudaLoop_->addCallback(buffer.stream, std::move(cudaCallback));
```
lw, line 231: I'm wondering whether we could call the user's callback here right away, after this line (rather than capture it in cudaCallback and thus invoke it some time later). The only guarantee we really give on those callbacks is that we invoke them when we've enqueued all the operations we need on the stream, and in fact after this line we've already done so.

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -0,0 +1,358 @@
...
+             << " is copying buffer from CUDA device to CPU";
+  auto tmpBuffer = makeCudaPinnedBuffer(buffer.length);
+  TP_CUDA_CHECK(cudaMemcpyAsync(
+      tmpBuffer.get(),
+      buffer.ptr,
+      buffer.length,
+      cudaMemcpyDeviceToHost,
+      buffer.stream));
+
+  auto cudaCallback =
```
lw, line 198: Shouldn't this be wrapped inside eagerCallbackWrapper? As it is, it only captures a _raw pointer_ to `this`, with no other means of keeping the Channel::Impl alive until the callback has fired. Which means the Channel::Impl could be destroyed (if the user closes it) and then this callback would cause a segfault when dereferencing `this`.

lw: Oh, never mind, I saw you added a TODO for this just a few lines down.

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -0,0 +1,358 @@
...
+            return;
+          }
+
+          CpuBuffer cpuBuffer{tmpBuffer.get(), buffer.length};
+          // Keep tmpBuffer alive until cpuChannel_ is done sending it over.
+          callback = eagerCallbackWrapper_(
+              [tmpBuffer{std::move(tmpBuffer)}, callback{std::move(callback)}](
+                  Impl& impl) { callback(impl.error_); });
+          cpuChannel_->send(
+              cpuBuffer, std::move(descriptorCallback), std::move(callback));
```
lw, line 228: Would be nice if this could be a separate method, since it's rather big to be a lambda.

To reuse the approach we're using in the pipe, it would be nice if we could have a method for each "reaction", i.e., each thing that the pipe/channel needs to do when some "signal" from outside comes in. This will also make it easier to convert this to an explicit state machine if we so wish.

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -0,0 +1,358 @@
...
+  const uint64_t sequenceNumber = nextTensorBeingReceived_++;
+  TP_VLOG(4) << "Channel " << id_ << " received a recv request (#"
+             << sequenceNumber << ")";
+
+  if (error_) {
+    callback(error_);
+    return;
+  }
+
+  TP_DCHECK_EQ(descriptor, std::string());
```
lw, line 269: Why would it be empty? It's the CPU channel's descriptor, it could be anything right?

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -0,0 +1,358 @@
...
+
+        // Keep tmpBuffer alive until cudaMemcpyAsync is done.
+        impl.cudaLoop_->addCallback(
+            buffer.stream,
+            [channelId{impl.id_},
+             tmpBuffer{std::move(tmpBuffer)}](cudaError_t cudaError) {
+              TP_VLOG(4) << "Channel " << channelId
+                         << " is done copying buffer from CPU to CUDA device";
+              TP_CUDA_CHECK(cudaError);
+            });
```
lw, line 300: Same here: could this all become a method?

lw (commented):
Looks good!

Most of my comment are on style issues (e.g., moving lambdas to methods), so feel free to ignore them if they're too pedantic.

The important comments are on a few "excessive" asserts (which could very well happen) and on the lifetime of some objects.

I haven't yet tried to really understand where the deadlock might come from, I'll look into that tomorrow.

tensorpipe/channel/cuda_basic/channel.cc
```diff
@@ -28,24 +28,12 @@ namespace tensorpipe {
...
-std::shared_ptr<uint8_t> makeCudaPinnedBuffer(size_t length) {
-  void* ptr;
-  TP_CUDA_CHECK(cudaMallocHost(&ptr, length));
-  return std::shared_ptr<uint8_t>(
-      reinterpret_cast<uint8_t*>(ptr),
-      [](uint8_t* ptr) { TP_CUDA_CHECK(cudaFreeHost(ptr)); });
-}
-
-} // namespace
-
```
beauby: This was moved to `common/cuda.h`.

tensorpipe/channel/cuda_basic/context.cc
```diff
@@ -47,13 +47,13 @@ class Context::Impl : public Context::PrivateIface,
   ~Impl() override = default;
 
  private:
-  std::string domainDescriptor_;
```
beauby: This is no longer stored, and instead we forward `cpuContext_->domainDescriptor()` where needed.

tensorpipe/channel/cuda_basic/context.cc
```diff
@@ -47,13 +47,13 @@ class Context::Impl : public Context::PrivateIface,
...
   std::atomic<bool> closed_{false};
   std::atomic<bool> joined_{false};
   ClosingEmitter closingEmitter_;
 
   std::shared_ptr<CpuContext> cpuContext_;
-  std::shared_ptr<CudaLoop> cudaLoop_;
+  // TODO: Lazy initialization of cuda loop.
+  CudaLoop cudaLoop_;
```
beauby: We can later make this an `std::optional` and lazy intialize it.

tensorpipe/common/cuda.h
```diff
@@ -8,14 +8,34 @@
...
+#define TP_CUDA_CHECK(a)                                                \
+  do {                                                                  \
+    cudaError_t error = (a);                                            \
+    TP_THROW_ASSERT_IF(cudaSuccess != error)                            \
+        << __TP_EXPAND_OPD(a) << " " << cudaGetErrorName(error) << " (" \
+        << cudaGetErrorString(error) << ")";                            \
+  } while (false)
+
+namespace tensorpipe {
+
```
beauby, line 27: We had mistakenly left those definitions in the global namespace. Fixing it here.

tensorpipe/common/cuda.h
```diff
@@ -8,14 +8,34 @@
...
 #include <tensorpipe/common/defs.h>
+#include <tensorpipe/common/error.h>
+
+#define TP_CUDA_CHECK(a)                                                \
+  do {                                                                  \
+    cudaError_t error = (a);                                            \
+    TP_THROW_ASSERT_IF(cudaSuccess != error)                            \
+        << __TP_EXPAND_OPD(a) << " " << cudaGetErrorName(error) << " (" \
+        << cudaGetErrorString(error) << ")";                            \
+  } while (false)
```
beauby, line 24: This definition is updated to avoid calling `cudaPeekAtLastError()`.

tensorpipe/common/cuda.h
```diff
@@ -8,14 +8,34 @@
...
+    return std::string(cudaGetErrorString(error_));
+  };
 
-#define TP_CUDA_CHECK(a)                                                      \
-  TP_THROW_ASSERT_IF(cudaSuccess != (a))                                      \
-      << __TP_EXPAND_OPD(a) << " " << cudaGetErrorName(cudaPeekAtLastError()) \
-      << " (" << cudaGetErrorString(cudaPeekAtLastError()) << ")"
+ private:
+  cudaError_t error_;
+};
```
beauby, line 38: This wraps a `cudaError_t` so that 1. we can have these set on the channel in case of failure rather than throw, and 2. we can call `CudaLoop::addCallback()` with a callback wrapped with `eagerCallbackWrapper_()`.

tensorpipe/common/cuda_loop.cc
```diff
@@ -13,66 +13,112 @@
...
+void CudaLoop::addCallback(
+    int device,
+    cudaStream_t stream,
+    std::function<void(const Error&)> callback) {
   std::unique_lock<std::mutex> lock(mutex_);
-  TP_THROW_ASSERT_IF(joined_);
+  if (closed_) {
+    callback(TP_CREATE_ERROR(CudaLoopClosedError));
+    return;
+  }
```
beauby, line 94: Calling callbacks in place with an error when the loop has already been closed.

tensorpipe/common/cuda_loop.cc
```diff
@@ -13,66 +13,112 @@
...
-        cv_.notify_all();
-        // This destroys the lambda, which is fine as long as no captured
-        // variable is being used passed that point.
-        cudaCallbacks_.erase(it);
-      };
-  TP_CUDA_CHECK(cudaStreamAddCallback(stream, runCudaCallback, &*it, 0));
+  auto cudaCallback =
+      std::make_unique<CudaCallback>(*this, std::move(callback));
+  lock.unlock();
+  TP_CUDA_CHECK(cudaSetDevice(device));
```
beauby, line 99: Setting the CUDA device here.

tensorpipe/common/cuda_loop.h
```diff
@@ -17,38 +18,46 @@
...
   std::thread thread_;
-  std::list<TCudaCallback> cudaCallbacks_;
   std::deque<Operation> operations_;
   std::mutex mutex_;
   std::condition_variable cv_;
-  uint64_t pendingOperations_ = 0;
-  bool joined_ = false;
+  uint64_t pendingOperations_{0};
+
+  bool closed_{false};
```
beauby, line 51: `closed_` is not an atomic because it is being guarded by `mutex_` in practice.

PR #248, opened 2020-12-10
Fix cuda_ipc channel for off set buffers.

beauby:
In the cuda_ipc channel, we use CUDA's IPC tools to export a buffer's address from one process to the other.
The `cudaIpcGetMemHandle()` function returns a handle to the _base address of the allocation_ of the given pointer, so we need to get the offset of the buffer from the base address and send that over to the other process.
That is only possible using the CUDA driver API, which means we have to link against it as well.

### DIFF
--- a/tensorpipe/channel/cuda_ipc/channel.cc
+++ b/tensorpipe/channel/cuda_ipc/channel.cc
@@ -80,7 +80,6 @@ class SendOperation {
     cudaIpcMemHandle_t handle;
     TP_CUDA_CHECK(cudaIpcGetMemHandle(&handle, const_cast<void*>(ptr_)));
     CUdeviceptr basePtr;
-
     CUresult error = cuMemGetAddressRange(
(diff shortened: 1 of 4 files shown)


tensorpipe/channel/cuda_ipc/channel.cc
```diff
@@ -77,9 +79,26 @@ class SendOperation {
...
+      CUresult res;
+      const char* errorName;
+      const char* errorStr;
+      res = cuGetErrorName(error, &errorName);
+      TP_THROW_ASSERT_IF(res != CUDA_SUCCESS);
+      res = cuGetErrorString(error, &errorStr);
+      TP_THROW_ASSERT_IF(res != CUDA_SUCCESS);
+
+      TP_THROW_ASSERT() << "CUDA error (" << errorName << "): " << errorStr;
+    }
```
beauby, line 96: This could be extracted to a helper function/macro, but since we're using the driver API only here for now I left it inline.

lw (approved):
Perfect!! I love this!

PR #253, opened 2020-12-16
Set device before creating CudaEvent from handle

mrshenli:
Although CUDA doc didn't mention that, it actually requires setting device when creating event from handle.

### DIFF
--- a/tensorpipe/test/channel/channel_test_cuda.cc
+++ b/tensorpipe/test/channel/channel_test_cuda.cc
@@ -236,6 +236,236 @@ class SendAcrossDevicesTest : public ClientServerChannelTestCase<CudaBuffer> {
 
 CHANNEL_TEST(CudaChannelTestSuite, SendAcrossDevices);
 
+class SendReverseAcrossDevicesTest
+    : public ClientServerChannelTestCase<CudaBuffer> {
+  static constexpr size_t kSize = 1024;
+
+ public:
+  void run(ChannelTestHelper<CudaBuffer>* helper) override {
+    if (TestEnvironment::numCudaDevices() < 2) {
+      GTEST_SKIP() << "Skipping test requiring >=2 CUDA devices.";
+    }
+
+    ClientServerChannelTestCase<CudaBuffer>::run(helper);
+  }
+
+ private:
+  void server(std::shared_ptr<transport::Connection> conn) override {
+    std::shared_ptr<CudaContext> ctx = this->helper_->makeContext("server");
+    auto channel = ctx->createChannel(std::move(conn), Endpoint::kListen);
+
+    // Send happens from device #1.
+    EXPECT_EQ(cudaSuccess, cudaSetDevice(1));
+    cudaStream_t sendStream;
+    EXPECT_EQ(cudaSuccess, cudaStreamCreate(&sendStream));
+    void* ptr;
+    EXPECT_EQ(cudaSuccess, cudaMalloc(&ptr, kSize));
+
+    // Set buffer to target value.
+    EXPECT_EQ(cudaSuccess, cudaMemsetAsync(ptr, 0x42, kSize, sendStream));
+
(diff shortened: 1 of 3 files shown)


tensorpipe/test/channel/channel_test_cuda.cc
```diff
@@ -236,6 +236,128 @@ class SendAcrossDevicesTest : public ClientServerChannelTestCase<CudaBuffer> {
...
+
+
+    // Perform send and wait for completion.
+    auto descriptorPromise = std::make_shared<
+        std::promise<std::tuple<tensorpipe::Error, std::string>>>();
+    auto sendPromise = std::make_shared<std::promise<tensorpipe::Error>>();
+    auto descriptorFuture = descriptorPromise->get_future();
+    auto sendFuture = sendPromise->get_future();
+
+    EXPECT_EQ(cudaSuccess, cudaSetDevice(0));
```
mrshenli, line 275: @beauby any reason `channel->send` and `channel->recv` must be called when properly setting the current device? This test would fail currently. But I we remove the two `EXPECT_EQ(cudaSuccess, cudaSetDevice(0));` it would pass.

mrshenli: please ignore the above comment. (it might be either I didn't recompile my test or the tests behavior is not consistent?) Currently, what I see is that, 0->0 and 0->1 both succeeded, but 1->1 and 1->0 failed. 

1->1 failed with the following error:

```
 ../tensorpipe/test/peer_group.h:140: Failure
Expected equality of these values:
  sizeof(len)
    Which is: 8
  ret
    Which is: 0
../tensorpipe/test/peer_group.h:49: Failure
Expected equality of these values:
  doneString_
    Which is: "done"
  recv(selfId)
    Which is: ""
../tensorpipe/test/peer_group.h:189: Failure
Value of: WIFEXITED(status)
  Actual: false
Expected: true
W1216 15:14:54.998178  3067 ../tensorpipe/test/peer_group.h:191] Peer process terminated with signal 11
../tensorpipe/test/peer_group.h:195: Failure
Expected equality of these values:
  0
  exitStatus
    Which is: 1
```

1->0 failed with the following error:

```
 1 FAILED TEST
../tensorpipe/test/peer_group.h:189: Failure
Value of: WIFEXITED(status)
  Actual: false
Expected: true
W1216 15:15:49.964588  3425 ../tensorpipe/test/peer_group.h:191] Peer process terminated with signal 11
../tensorpipe/test/peer_group.h:195: Failure
Expected equality of these values:
  0
  exitStatus
    Which is: 1
```

mrshenli: got a little different output when testing 1->1, the `std::bad_alloc` looks suspicious:

```
../tensorpipe/test/peer_group.h:140: Failure
Expected equality of these values:
  sizeof(len)
    Which is: 8
  ret
    Which is: 0
../tensorpipe/test/peer_group.h:189: Failure
Value of: WIFEXITED(status)
  Actual: false
Expected: true
W1216 15:25:08.701485  5264 ../tensorpipe/test/peer_group.h:191] Peer process terminated with signal 11
unknown file: Failure
C++ exception with description "std::bad_alloc" thrown in the test body.
```

tensorpipe/test/channel/channel_test_cuda.cc
```diff
@@ -236,6 +236,236 @@ class SendAcrossDevicesTest : public ClientServerChannelTestCase<CudaBuffer> {
 
 CHANNEL_TEST(CudaChannelTestSuite, SendAcrossDevices);
 
+class SendReverseAcrossDevicesTest
```
mrshenli, line 239: these two tests are mostly duplicated from the above test, and the only differences are src/dst devices.

beauby (approved):

PR #259, opened 2021-01-11
Reuse pinned buffers in cuda_basic.

beauby:
This PR introduces a `CudaPinnedMemoryAllocator` (shared among
`cuda_basic` channels from a same context) that pre-allocates one
large slab of host pinned memory, divided into chunks that can be
used by channels when they need to stage data to host buffers.

The proposed strategy is to return a whole existing chunk when one is
available and the requested buffer is at most the chunk
size. Otherwise, we fall back to allocating a new temporary pinned
buffer.
We could consider copying a large device tensor to multiple
non-contiguous chunks but it is not clear there would be any perf
benefits of exchanging one `cudaMallocHost()` call for multiple
`cudaMemcpyAsync()`.
~~One optimization that would make sense would be to use several
contiguous chunks when possible.~~ done.

### DIFF
--- a/tensorpipe/common/cuda.h
+++ b/tensorpipe/common/cuda.h
@@ -9,6 +9,8 @@
 #pragma once
 
 #include <memory>
+#include <mutex>
+#include <vector>
 
 #include <cuda_runtime.h>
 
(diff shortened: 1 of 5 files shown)


tensorpipe/common/cuda.h
```diff
@@ -125,14 +127,67 @@ inline int cudaDeviceForPointer(const void* ptr) {
...
-using CudaPinnedBuffer = std::shared_ptr<uint8_t>;
+class CudaPinnedBufferAllocator {
+ public:
+  CudaPinnedBufferAllocator(const CudaPinnedBufferAllocator&) = delete;
+  CudaPinnedBufferAllocator(CudaPinnedBufferAllocator&&) = delete;
+  CudaPinnedBufferAllocator& operator=(const CudaPinnedBufferAllocator&) =
+      delete;
+  CudaPinnedBufferAllocator& operator=(CudaPinnedBufferAllocator&&) = delete;
+
+  CudaPinnedBufferAllocator(size_t numChunks = 1024, size_t chunkSize = 1024)
```
beauby, line 138: These defaults are arbitrary for now. We should figure out sensible values, and expose those parameters as options to the context.

tensorpipe/common/cuda.h
```diff
@@ -125,14 +127,66 @@ inline int cudaDeviceForPointer(const void* ptr) {
...
+          }
+
+          return std::shared_ptr<uint8_t>(
+              basePtr_.get() + startChunk * kChunkSize_,
+              [this, startChunk, curChunk](uint8_t* /* unused */) {
+                for (size_t i = startChunk; i <= curChunk; ++i) {
+                  // There is a harmless race condition here.
+                  chunkAvailable_[i] = true;
+                }
+              });
```
beauby, line 166: The assumption here is that the allocator will not be destroyed until all channels in the context have invoked all their callbacks, so every chunk will have been reclaimed by then.

tensorpipe/common/cuda.h
```diff
@@ -125,14 +127,66 @@ inline int cudaDeviceForPointer(const void* ptr) {
...
+    {
+      std::unique_lock<std::mutex> lock(mutex_);
+      size_t requestedChunks = (size + kChunkSize_ - 1) / kChunkSize_;
+      size_t startChunk = 0;
+
+      // Linear check for available contiguous chunks.
+      for (size_t curChunk = 0; curChunk < kNumChunks_; ++curChunk) {
+        if (!chunkAvailable_[curChunk]) {
+          startChunk = curChunk + 1;
+        } else if (curChunk - startChunk + 1 == requestedChunks) {
```
lw, line 154: This means we'll only use the pool when the whole chunk fits contiguously? Hence any tensor that is larger than the pool will always cause an ad-hoc allocation, right? And that due to fragmentation also tensors that might fit in the pool could cause an allocation if they are unlucky?

tensorpipe/common/cuda.h
```diff
@@ -125,14 +127,66 @@ inline int cudaDeviceForPointer(const void* ptr) {
...
-inline CudaPinnedBuffer makeCudaPinnedBuffer(size_t length) {
-  void* ptr;
-  TP_CUDA_CHECK(cudaMallocHost(&ptr, length));
-  return CudaPinnedBuffer(reinterpret_cast<uint8_t*>(ptr), [](uint8_t* ptr) {
-    TP_CUDA_CHECK(cudaFreeHost(ptr));
-  });
-}
+    // No contiguous pre-allocated chunks available, falling back to allocating
+    // a temporary buffer.
+    return allocPinnedBuffer(size);
```
lw, line 173: I'm worried that this may be problematic, as it means that the original problem we're trying to solve could still occur, less frequently but more unpredictably.

You said that one tradeoff you considered is between the cost of one cudaHostMalloc call and multiple cudaMemcpyAsync calls, and I guess you deemed the latter to be more undesirable than the former. I however am leaning the other way, as a cudaHostMalloc delays _all_ the kernels that the process is launching, even those that are completely unrelated to TP. In other words, it completely breaks any computation/communication overlap the user might be trying to do.

On the other hand, I think we can mitigate and amortize the cost of cudaMemcpyAsync by having a large enough buffer and "pipelining" correctly: since I expect the bottleneck to be TCP (since PCIe should hopefully be faster), we should have some slack in our PCIe transfers and thus we should be able to "hide" some overhead.

I also believe that we shouldn't worry too much about delaying some transfers when the buffer is full (i.e., having those channels wait for a slot to free up, rather than allocating new memory). That's because if the buffer is full it means that there already are many transfers happening, and thus even if we immediately added an additional one it may not receive bandwidth until the previous ones finish.

This clearly requires a rework of the channel logic, in order for it to request a buffer and deal with the possibility of it not being available right away, and thus giving a callback in order to be woken up later.

lw (changes requested):

beauby:
Closing in favor of #303 (comments were addressed).

PR #260, opened 2021-01-11
Use docker for GPU builds to test against multiple CUDA versions

beauby:
Adding CI jobs with CUDA 9.2, 10.1, 10.2, 11.0, and 11.1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,46 +62,39 @@ jobs:
             cd build
             make install
   build_gpu:
+    parameters:
+      cuda_version:
+        type: string
     machine:
       resource_class: gpu.nvidia.small
       image: ubuntu-1604-cuda-10.1:201909-23
+      docker_layer_caching: true
     steps:
       - checkout
-      - run:
-          name: Install apt packages
-          command: |
-            sudo apt-get update
-            sudo apt-get install -y git-core build-essential cmake
       - run:
           name: Initialize submodules
           command: |
             git submodule init
             git submodule update
       - run:
-          name: Build
+          name: Build/test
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -62,46 +62,39 @@ jobs:
...
-          command: |
-            cd build
-            sudo make install
+              docker build -t tensorpipe -f .circleci/Dockerfile.cuda<< parameters.cuda_version >> .
+              docker run --gpus all tensorpipe sh -c "
+                mkdir build && cd build &&
+                cmake ../ \
+                  -DCMAKE_C_FLAGS=\"-Werror -Wno-deprecated-declarations\" \
+                  -DCMAKE_CXX_FLAGS=\"-Werror -Wno-deprecated-declarations\" \
+                  -DCUDA_NVCC_FLAGS=\"-gencode arch=compute_61,code=sm_61\" \
```
lw, line 88: This line is new, why is it needed?

lw (approved):
Very nice! Thanks!

Could we also add CUDA 9.2 and 11.0? They are two of the versions supported by PyTorch (as of v1.7, not sure if this will change in v1.8) so I think we should cover them too.

.circleci/config.yml
beauby: It is not needed strictly speaking, but it allows for generating device code optimized for the GPU type available in CircleCI instances.

lw:
The clang-format issue has been fixed in master, the TSAN one seems unrelated and I've opened https://github.com/pytorch/tensorpipe/issues/262 to track it.

PR #263, opened 2021-01-12
Fix for cudaPointerGetAttributes returning invalid value.

beauby:


lw (commented):
I'm still puzzled by this bug. I don't see why setting the device here would change anything.

I tried to reproduce the steps of this code in Python:
```
In [1]: import ctypes

In [2]: cuda = ctypes.cdll.LoadLibrary("libcudart.so")

In [6]: dev = ctypes.c_int(9)

In [7]: cuda.cudaGetDevice(ctypes.byref(dev))
Out[7]: 0

In [8]: dev
Out[8]: c_int(0)

In [9]: cuda.cudaSetDevice(dev)
Out[9]: 0
```
and after this, with the IPython prompt still open, I ran nvidia-smi, but I didn't see any process with an open CUDA context on GPU 0.

beauby:
Closing this as we merged an other solution to this problem.

PR #265, opened 2021-01-13
Remove unnecessary CudaDeviceGuard for cudaIpcOpenEventHandle.

beauby:
As hinted by the (official) simpleIPC CUDA example, calling `cudaSetDevice()` before opening an IPC event handle is not necessary (cf. https://github.com/NVIDIA/cuda-samples/blob/master/Samples/simpleIPC/simpleIPC.cu#L126). Removing the extraneous device guard.


beauby:
Closing as we moved to using low-overhead `DeviceGuard`s everywhere.

PR #266, opened 2021-01-13
Mark cuda_ipc viable only if all devices have peer access available.

beauby:
Since `cudaIpcOpenMemHandle()` will lazily try to enable peer access,
we must ensure it will not fail.
One limitation of this check is that it can be fooled by the
`CUDA_VISIBLE_DEVICES` environment variable, as two processes might be
each limited to a viable set of devices, while pairs across those sets
might not be able to access each other. One option would be to use an
ordered list of device UUIDS as part of the domain descriptor (those
would need to be accessed through the `nvmlDeviceGetUUID()` function
from the nvml library, or through the PCI ids returned in `cudaDeviceGetAttribute()`). A simpler option would be to check the
process' environment and mark the channel non-viable if
`CUDA_VISIBLE_DEVICES` is set, or to add the value of 
`CUDA_VISIBLE_DEVICES` to the domain descriptor.

### DIFF
--- a/tensorpipe/channel/cuda_ipc/context_impl.cc
+++ b/tensorpipe/channel/cuda_ipc/context_impl.cc
@@ -58,12 +58,15 @@ bool ContextImpl::isViable() const {
     return false;
   }
 
+  // This part is largely inspired from
+  // https://github.com/NVIDIA/cuda-samples/blob/master/Samples/simpleIPC/simpleIPC.cu.
   int deviceCount;
   TP_CUDA_CHECK(cudaGetDeviceCount(&deviceCount));
   for (int i = 0; i < deviceCount; ++i) {
     cudaDeviceProp props;
     TP_CUDA_CHECK(cudaGetDeviceProperties(&props, i));
 
+    // Unified addressing is required for IPC.
     if (!props.unifiedAddressing) {
       TP_VLOG(4) << "Channel context " << id_
                  << " is not viable because CUDA device " << i
@@ -71,6 +74,8 @@ bool ContextImpl::isViable() const {
       return false;
     }
 
+    // The other two compute modes are "exclusive" and "prohibited", both of
+    // which prevent access from an other process.
     if (!props.computeMode != cudaComputeModeDefault) {
       TP_VLOG(4) << "Channel context " << id_
                  << " is not viable because CUDA device " << i
@@ -79,6 +84,12 @@ bool ContextImpl::isViable() const {
     }
 
     for (int j = 0; j < deviceCount; ++j) {
+      // cudaDeviceCanAccessPeer() returns false when the two devices are the
+      // same.
+      if (i == j) {
+        continue;
+      }
+
       int canAccessPeer;
       TP_CUDA_CHECK(cudaDeviceCanAccessPeer(&canAccessPeer, i, j));
       if (!canAccessPeer) {
--- a/tensorpipe/channel/cuda_ipc/context_impl.cc
+++ b/tensorpipe/channel/cuda_ipc/context_impl.cc
@@ -76,7 +76,7 @@ bool ContextImpl::isViable() const {
 
     // The other two compute modes are "exclusive" and "prohibited", both of
     // which prevent access from an other process.
-    if (!props.computeMode != cudaComputeModeDefault) {
+    if (props.computeMode != cudaComputeModeDefault) {
       TP_VLOG(4) << "Channel context " << id_
                  << " is not viable because CUDA device " << i
                  << " is not in default compute mode";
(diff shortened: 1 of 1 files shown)


lw:
Didn't look at the code yet, but here are a few comments on the summary.

I agree that we should find a way to deal with CUDA_VISIBLE_DEVICES, for the reason you mentioned. The ideal solution would be to somehow sidestep CUDA_VISIBLE_DEVICES so that we can ensure there's P2P between all the devices on the machine, even those we cannot see. It appears that NVML can do that, i.e., it does not honor CUDA_VISIBLE_DEVICES. (I'm saying that because I ran `CUDA_VISIBLE_DEVICES=0 nvidia-smi` and it still displayed all GPUs). NVML then has the `nvmlDeviceGetP2PStatus` function that would return us the information we need. However, I don't know if it's a good idea to start playing with a third library (in addition to runtime and driver) a few weeks before the release. We'll certainly hit issues. So I'd say let's look for a temporary solution for the release. Doing what you suggests seems very reasonable to me: let's append the (sorted) UUIDs of the devices to the descriptor, so that if the two endpoints don't see the same devices they'll refuse to use CUDA_IPC and they will fall back onto CUDA_BASIC. Note that the UUID can be retrieved through CUDA (no need for NVML), through the `cudaGetDeviceProperties` call.

tensorpipe/channel/cuda_ipc/context_impl.cc
```diff
@@ -55,7 +52,45 @@ std::shared_ptr<CudaChannel> ContextImpl::createChannel(
...
+    if (!props.computeMode != cudaComputeModeDefault) {
+      TP_VLOG(4) << "Channel context " << id_
+                 << " is not viable because CUDA device " << i
+                 << " is not in default compute mode";
+      return false;
+    }
+
+    for (int j = 0; j < deviceCount; ++j) {
+      int canAccessPeer;
+      TP_CUDA_CHECK(cudaDeviceCanAccessPeer(&canAccessPeer, i, j));
```
lw, line 83: Does this behave correctly when called with `i == j`?

tensorpipe/channel/cuda_ipc/context_impl.cc
```diff
@@ -55,7 +52,45 @@ std::shared_ptr<CudaChannel> ContextImpl::createChannel(
...
+                 << " does not have unified addressing";
+      return false;
+    }
+
+    if (!props.computeMode != cudaComputeModeDefault) {
+      TP_VLOG(4) << "Channel context " << id_
+                 << " is not viable because CUDA device " << i
+                 << " is not in default compute mode";
+      return false;
+    }
```
lw, line 79: Why does this matter? What happens if that's not the case?

tensorpipe/channel/cuda_ipc/context_impl.cc
```diff
@@ -55,7 +52,45 @@ std::shared_ptr<CudaChannel> ContextImpl::createChannel(
...
+    return false;
+  }
+
+  int deviceCount;
+  TP_CUDA_CHECK(cudaGetDeviceCount(&deviceCount));
+  for (int i = 0; i < deviceCount; ++i) {
+    cudaDeviceProp props;
+    TP_CUDA_CHECK(cudaGetDeviceProperties(&props, i));
+
+    if (!props.unifiedAddressing) {
```
lw, line 67: I remember we needed unified addressing, but I already forgot exactly why. Could you put a comment explaining it?

tensorpipe/channel/cuda_ipc/context_impl.cc
beauby: Good catch, it looks like it does not. We should add tests to check viability of channels in channel tests.

tensorpipe/channel/cuda_ipc/context_impl.cc
beauby: The other two modes are "exclusive" and "prohibited", both of which prevent the device to be accessed by an other process (cf. https://github.com/NVIDIA/cuda-samples/blob/master/Samples/simpleIPC/simpleIPC.cu#L219).

tensorpipe/channel/cuda_ipc/context_impl.cc
beauby: According to the docs:
> IPC functionality is restricted to devices with support for unified addressing on Linux operating systems.

tensorpipe/channel/cuda_ipc/context_impl.cc
lw: Ok, interesting. Could you add a comment in the code with this information? Linking to that demo seems also a good idea.

tensorpipe/channel/cuda_ipc/context_impl.cc
lw: Could you put this in the code, as a comment, so we'll remember it next time we read this?

mrshenli (approved):
Stamp to unblock, but I am not confident about the impact. :)

PR #267, opened 2021-01-13
Fix cudaDeviceForPointer when called on uninitialized device

lw:
Differential Revision: D25902324

### DIFF
--- a/tensorpipe/common/cuda.h
+++ b/tensorpipe/common/cuda.h
@@ -126,11 +126,12 @@ inline int cudaDeviceForPointer(const void* ptr) {
   return attrs.device;
 }
 
-inline int cudaFixedDeviceForPointer(CudaLib& cudaLib, const void* ptr) {
+inline int cudaFixedDeviceForPointer(const CudaLib& cudaLib, const void* ptr) {
   CUcontext ctx;
-  TP_CUDA_DRIVER_CHECK(cudaLib, cudaLib.ctxPopCurrent(&ctx));
+  TP_CUDA_DRIVER_CHECK(cudaLib, cudaLib.ctxGetCurrent(&ctx));
+  TP_CUDA_DRIVER_CHECK(cudaLib, cudaLib.ctxSetCurrent(nullptr));
   int deviceIdx = cudaDeviceForPointer(ptr);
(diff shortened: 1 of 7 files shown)


tensorpipe/common/cuda.h
```diff
@@ -125,6 +126,15 @@ inline int cudaDeviceForPointer(const void* ptr) {
   return attrs.device;
 }
 
+inline int cudaFixedDeviceForPointer(const CudaLib& cudaLib, const void* ptr) {
+  CUcontext ctx;
+  TP_CUDA_DRIVER_CHECK(cudaLib, cudaLib.ctxGetCurrent(&ctx));
+  TP_CUDA_DRIVER_CHECK(cudaLib, cudaLib.ctxSetCurrent(nullptr));
+  int deviceIdx = cudaDeviceForPointer(ptr);
+  TP_CUDA_DRIVER_CHECK(cudaLib, cudaLib.ctxSetCurrent(ctx));
```
beauby, line 134: Out of curiosity, isn't this effctively equivalent to the following? 
```
cudaCtxPopCurrent(&ctx);
int deviceIdx = cudaDeviceForPointer(ptr);
cudaCtxPushCurrent(ctx);
```

tensorpipe/test/common/cuda_test.cc
```diff
@@ -0,0 +1,80 @@
...
+        },
+        [&]() {
+          std::string ptrStr = pg.recv(PeerGroup::kClient);
+          void* ptr = *reinterpret_cast<void**>(&ptrStr[0]);
+
+          EXPECT_EQ(tensorpipe::cudaFixedDeviceForPointer(getCudaLib(), ptr), 1);
+        });
+  }
+
+  {
```
beauby, line 56: Could we split out those into two separate tests?

tensorpipe/common/cuda.h
lw: Surprisingly it isn't. I can't say why, but that's actually what I tried in the first version of this PR (https://github.com/pytorch/tensorpipe/pull/267/commits/b5ee2b5e4b6f6b8d9111e30b25eb5be8bdade040) and you can see that it caused a deadlock and another failure.

PR #269, opened 2021-01-18
Run CI on 2-GPU instances.

beauby:


.circleci/config.yml
```diff
@@ -93,7 +93,7 @@ jobs:
                   -DTP_ENABLE_IBV=OFF \
                   -DTP_BUILD_TESTING=ON &&
                 make -j20 &&
-                ./tensorpipe/test/tensorpipe_test --gtest_filter=-*CudaGdr* &&
+                ./tensorpipe/test/tensorpipe_test --gtest_filter=-*CudaGdr*:CudaIpc/CudaChannelTestSuite.*Across*Devices* &&
```
lw, line 96: I wonder if we should spell out each of these test names in full, to avoid like someone adding a test called "DoNotSendAcrossDevices" which would accidentally be skipped by this without them knowing. It's clearly unlikely, but you get the idea. If we do that though we might want to find a way to make this list more readable. Like, is there a way to tell gtest to load the skiplist from a file?

lw: (Note this is a nit, no action required, we can deal with it later)

PR #273, opened 2021-01-19
Ensure tensor order in cuda_basic.

beauby:
The CudaLoop does not guarantee a FIFO order of callbacks, therefore delaying the `cudaMemcpyAsync()`to the `CudaLoop` can lead to an inversion in the order in which tensors are sent through the CPU channel . This change ensures the tensors are sent through the CPU channel in the same order they are being sent through the wrapping cuda_basic channel.


mrshenli (approved):
LGTM! Thanks for super prompt fix!!

lw (approved):
Looks good. How was this tested?

beauby:
@lw Manually for now, but you have a pending test for this, right?

PR #275, opened 2021-01-19
Add test for CudaXth with P2P disabled.

beauby:


beauby:
As it turns out, P2P is disabled by default, so this test is useless (it is actually failing because it tries to disable P2P while it is not enabled).

PR #280, opened 2021-01-20
Add viability check to cuda_xth.

beauby:
Universal addressing is required for cross-device `cudaMemcpyAsync()`,
so this check ensures all visible devices have UVA enabled. This is
the case for all recent devices (i.e. compute capability >=2.0) on
64bit processes.
Should we want to support 32bit processes or very old
hardware (pre 2011 from what I gathered), we could add a fallback to
`cudaMemcpyPeerAsync()` which supports passing the devices as parameters.

### DIFF
--- a/tensorpipe/channel/cuda_xth/context_impl.cc
+++ b/tensorpipe/channel/cuda_xth/context_impl.cc
@@ -16,6 +16,7 @@
 #include <utility>
 
 #include <tensorpipe/channel/cuda_xth/channel_impl.h>
+#include <tensorpipe/common/cuda.h>
 #include <tensorpipe/common/defs.h>
 #include <tensorpipe/common/optional.h>
 #include <tensorpipe/common/system.h>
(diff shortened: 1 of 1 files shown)


tensorpipe/channel/cuda_xth/context_impl.cc
```diff
@@ -62,7 +63,28 @@ std::shared_ptr<CudaChannel> ContextImpl::createChannel(
...
+
+  int deviceCount;
+  TP_CUDA_CHECK(cudaGetDeviceCount(&deviceCount));
+  for (int i = 0; i < deviceCount; ++i) {
+    cudaDeviceProp props;
+    TP_CUDA_CHECK(cudaGetDeviceProperties(&props, i));
+
+    // Unified addressing is required for cross-device `cudaMemcpyAsync()`. We
+    // could lift this requirement by adding a fallback to
+    // `cudaMemcpyPeerAsync()`.
```
lw, line 78: If you have a reference it would be nice to link to it here. Otherwise it's fine too, unified addressing seems common enough that there's no real need to justify it.

lw (approved):

PR #281, opened 2021-01-20
Add device uuids to domain descriptor of CudaIpc.

beauby:
This ensures that whenever the domain descriptor matches and the
channel is viable, it will indeed work, avoiding the current loophole
where a machine could have two sets of pairwise p2p-enabled devices
with no p2p across the sets, fooling our viability check if each
process is restricted to one of the sets.

### DIFF
--- a/tensorpipe/channel/cuda_ipc/context_impl.cc
+++ b/tensorpipe/channel/cuda_ipc/context_impl.cc
@@ -36,6 +36,8 @@ std::string deviceUUID(int device) {
 #if CUDART_VERSION > 9020
   uuid << std::setfill('0') << std::setw(2);
   for (int j = 0; j < 16; ++j) {
+    // The bitmask is required otherwise a negative value will get promoted to
+    // (signed) int with sign extension if char is signed.
     uuid << (props.uuid.bytes[j] & 0xff);
   }
 #else // CUDART_VERSION <= 9020
--- a/tensorpipe/channel/cuda_ipc/context_impl.cc
+++ b/tensorpipe/channel/cuda_ipc/context_impl.cc
@@ -15,6 +15,7 @@
 #include <utility>
 
 #include <tensorpipe/channel/cuda_ipc/channel_impl.h>
+#include <tensorpipe/common/cuda.h>
 #include <tensorpipe/common/defs.h>
 #include <tensorpipe/common/optional.h>
 #include <tensorpipe/common/system.h>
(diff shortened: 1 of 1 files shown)


tensorpipe/channel/cuda_ipc/context_impl.cc
```diff
@@ -30,6 +31,26 @@ std::string generateDomainDescriptor() {
...
+  int deviceCount;
+  TP_CUDA_CHECK(cudaGetDeviceCount(&deviceCount));
+  std::vector<std::string> uuids(deviceCount);
+  for (int i = 0; i < deviceCount; ++i) {
+    cudaDeviceProp props;
+    TP_CUDA_CHECK(cudaGetDeviceProperties(&props, i));
+    std::ostringstream uuid;
+    uuid << std::hex << std::setw(2) << std::setfill('0');
+    for (int j = 0; j < 16; ++j) {
+      uuid << (props.uuid.bytes[j] & 0xff);
```
beauby, line 44: The `& 0xff` is needed because `char` is apparently signed, and `props.uuid.bytes[j]` gets promoted to (signed) `int` with sign-extension.

tensorpipe/channel/cuda_ipc/context_impl.cc
```diff
@@ -25,11 +26,46 @@ namespace cuda_ipc {
...
+  uuid << std::hex;
+
+#if CUDART_VERSION > 9020
+  uuid << std::setfill('0') << std::setw(2);
+  for (int j = 0; j < 16; ++j) {
+    uuid << (props.uuid.bytes[j] & 0xff);
+  }
+#else // CUDART_VERSION <= 9020
+  // CUDA <=9.2 does not have a uuid device property. Falling back to using PCI
+  // ids.
```
lw, line 43: The CUDA driver seems to have a function for that though also in version 9.2.

https://docs.nvidia.com/cuda/archive/9.2/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g987b46b884c101ed5be414ab4d9e60e4

tensorpipe/channel/cuda_ipc/context_impl.cc
lw: As a general rule, this type of remarks are better put as comments in the code itself, since they will be helpful to future readers too, which won't see this review thread. My philosophy on this is: comments about why the *new* code, in and of itself, is a certain way (i.e., the *outcome* of the change) go into the code; comments about the *change*, that explain the difference between the old and the new version (and thus require context on both) go into the review thread.

lw: As for the change itself, I wonder if we can't just reinterpret_cast this (or even static_cast it? not sure) to an unsigned char. Would it be more readable?

lw: One more nit (feel free to ignore): nvidia-smi seems to represent the UUID in the "standard" format with sections broken up by dashes (4-2-2-2-6 bytes). It might be nice to match this for easier comparison.

lw (approved):

PR #287, opened 2021-02-01
Rewrite domainDescriptor test.

beauby:
* #303 Optimize pinned memory allocation in CudaBasic.
* #302 Fix synchronization in CUDA tests.
* #289 Add control connection for CudaBasic.
* #288 Refactor channel client-server tests.
* **#287 Rewrite domainDescriptor test.**

Differential Revision: [D26514270](https://our.internmc.facebook.com/intern/diff/D26514270)


tensorpipe/test/channel/channel_test.cc
```diff
@@ -17,23 +17,21 @@ using namespace tensorpipe::channel;
...
+    auto peerGroup = helper->makePeerGroup();
+    peerGroup->spawn(
+        [&] {
+          std::shared_ptr<Context<TBuffer>> context1 =
+              helper->makeContext("ctx1");
+          std::shared_ptr<Context<TBuffer>> context2 =
+              helper->makeContext("ctx2");
+          EXPECT_FALSE(context1->domainDescriptor().empty());
+          EXPECT_FALSE(context2->domainDescriptor().empty());
+          EXPECT_EQ(context1->domainDescriptor(), context2->domainDescriptor());
```
lw, line 32: Just noticed (and unrelated to this diff): do we want to now change this to check that `context1->canCommunicateWithRemote(context2->domainDescriptor())`?

tensorpipe/test/channel/channel_test.cc
```diff
@@ -17,23 +17,21 @@ using namespace tensorpipe::channel;
...
+        },
+        [] {});
```
lw, line 34: I see you removed the `done` and `join` calls. I don't remember what they were for. Did they only ensure there was a "barrier"? (in which case I agree we don't need them)

lw (approved):

tensorpipe/test/channel/channel_test.cc
beauby: Yeah so join and done are just a messaging-based barrier on the peer group. The peer group itself joins both threads/processes, but the barrier is/was needed to keep objects alive on the side that would join first.

tensorpipe/test/channel/channel_test.cc
beauby: Good question. In its current form, this test somewhat ensures that the domain descriptor is the same in subsequent calls. I'm not sure which property we want to test.

lw: Well we can't really check for any "deep" semantic meaning of the descriptor (like, that it compares true or false in the various scenarios), as that varies for each channel and it requires specific test setups. So I guess we're mainly checking for trivial properties, and "reflexivity" seems a good one. However, I also now think that ensuring that all contexts started in the same conditions should produce the same descriptor. So maybe we should both check for equality, and that `canCommunicateWithRemote` is true?

PR #288, opened 2021-02-01
Refactor channel client-server tests.

beauby:
* #303 Optimize pinned memory allocation in CudaBasic.
* #302 Fix synchronization in CUDA tests.
* #289 Add control connection for CudaBasic.
* **#288 Refactor channel client-server tests.**
* #287 Rewrite domainDescriptor test.

Client/server channel tests now deal only with a pre-established
channel. This allows us to automatically create the required number of
connections per channel.
In order to guarantee the connection order on each side, the client
side send the connection id over the connection so that the server
side can match it properly.

Differential Revision: [D26514269](https://our.internmc.facebook.com/intern/diff/D26514269)


tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
 
 template <typename TBuffer>
 class ClientServerChannelTestCase : public ChannelTestCase<TBuffer> {
+  std::vector<std::future<std::shared_ptr<tensorpipe::transport::Connection>>>
+  accept(
+      std::shared_ptr<tensorpipe::transport::Listener> listener,
```
lw, line 208: Nit: the listener is only needed to issue the `accept` call, hence this function doesn't need a _copy_ of the shared_ptr that holds the listener, it could simply take a reference (in fact, a const reference). However, in turn, it would be odd to take a reference to a (smart) pointer (imagine seeing `void foo(uint8_t* const& data)`), it would make more sense to get a reference to the value pointed to by the pointer, rather than the pointer itself. Hence, to sum up, I think the most "idiomatic" way to pass this in here is `tensorpipe::transport::Listener& listener`. This has the benefit that it doesn't mandate how the lifetime of the listener is managed (could be a shared_ptr, a unique_ptr, stack-allocated, ...).

I'm being somewhat pedantic about it because I had spent some time thinking about it and wondering what's the best approach, and wanted to shared my thoughts. This is also explained much better than I could do here: https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
...
+      size_t numConnections) {
+    auto promises = std::make_shared<std::vector<
+        std::promise<std::shared_ptr<tensorpipe::transport::Connection>>>>(
+        numConnections);
+    for (size_t i = 0; i < numConnections; ++i) {
+      listener->accept(
+          [&, promises](
```
lw, line 215: What are we capturing by reference here? It seems somewhat dangerous, could we avoid it?

tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
...
+              std::shared_ptr<tensorpipe::transport::Connection> connection) {
+            ASSERT_FALSE(error) << error.what();
+            connection->read([promises, connection](
+                                 const tensorpipe::Error& error,
+                                 const void* connIdBuf,
+                                 size_t length) {
+              ASSERT_FALSE(error) << error.what();
+              ASSERT_EQ(sizeof(uint64_t), length);
+              promises->at(*static_cast<const uint64_t*>(connIdBuf))
+                  .set_value(std::move(connection));
```
lw, line 226: If you're moving then the lambda needs to be mutable.

tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
...
+    auto promises = std::make_shared<std::vector<
+        std::promise<std::shared_ptr<tensorpipe::transport::Connection>>>>(
+        numConnections);
+    for (size_t i = 0; i < numConnections; ++i) {
+      listener->accept(
+          [&, promises](
+              const tensorpipe::Error& error,
+              std::shared_ptr<tensorpipe::transport::Connection> connection) {
+            ASSERT_FALSE(error) << error.what();
+            connection->read([promises, connection](
```
lw, line 219: Nit: you could move `promises` into the lambda.

tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
...
+  std::vector<std::shared_ptr<tensorpipe::transport::Connection>> connect(
+      std::shared_ptr<tensorpipe::transport::Context> transportCtx,
+      std::string addr,
+      size_t numConnections) {
+    std::vector<std::shared_ptr<tensorpipe::transport::Connection>> connections(
+        numConnections);
+    for (size_t connId = 0; connId < numConnections; ++connId) {
+      connections[connId] = transportCtx->connect(addr);
+      auto connIdBuf = std::make_shared<uint64_t>();
+      *connIdBuf = connId;
```
lw, line 250:
```suggestion
      auto connIdBuf = std::make_shared<uint64_t>(connId);
```
Wondering if this would work?

tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
...
+      size_t numConnections) {
+    auto promises = std::make_shared<std::vector<
+        std::promise<std::shared_ptr<tensorpipe::transport::Connection>>>>(
+        numConnections);
+    for (size_t i = 0; i < numConnections; ++i) {
+      listener->accept(
+          [&, promises](
+              const tensorpipe::Error& error,
+              std::shared_ptr<tensorpipe::transport::Connection> connection) {
+            ASSERT_FALSE(error) << error.what();
```
lw, line 218: Should we set the corresponding promise to error in this case, so that the main thread doesn't get blocked forever?

tensorpipe/test/channel/channel_test.h
```diff
@@ -202,6 +203,62 @@ class ChannelTestCase {
...
+            connection->read([promises, connection](
+                                 const tensorpipe::Error& error,
+                                 const void* connIdBuf,
+                                 size_t length) {
+              ASSERT_FALSE(error) << error.what();
```
lw, line 223: Same here, should we set promise as errored?

lw (approved):
Looks really good! Thanks! I've only left minor style nits

tensorpipe/test/channel/channel_test.h
beauby: This was a remnant of the past, removed.

tensorpipe/test/channel/channel_test.h
beauby: Does it? I'm just moving an argument here, not a captured variable.

tensorpipe/test/channel/channel_test.h
beauby: The `promises` vector needs to be shared among all lambdas. Am I missing something?

lw: A copy of the shared_ptr is already made by the closure of the first lambda (line 215), hence this shared_ptr here will not need to be copied further, and that's why it can be moved.

tensorpipe/test/channel/channel_test.h
lw: The arguments of this lambda are `error`, `connIdBuf` and `length`. The `connection` was an argument to the "parent" lambda, but it has since been captured into this child lambda (see line 219).

tensorpipe/test/channel/channel_test.h
beauby: Making the promise error would just delay the throwing of the exception, no? What did you have in mind?

lw: Would it? My worry is that on line 284 we call `get()` on the future, and if we throw the error here then this thread crashes (which could be some event loop), and thus the future will never be completed, and thus the main thread on like 284 will hang forever.

tensorpipe/test/channel/channel_test.h
```diff
@@ -203,39 +203,51 @@ class ChannelTestCase {
...
-        futures;
-
-    for (auto& p : *promises) {
-      futures.push_back(p.get_future());
+      listener.accept([promise, connections, connectionsMutex](
+                          const tensorpipe::Error& error,
+                          std::shared_ptr<tensorpipe::transport::Connection>
+                              connection) mutable {
+        if (error) {
+          promise->set_value(std::make_pair(error, *connections));
```
lw, line 223: This risks calling the `set_value` method multiple times, which is an error and would throw. It could happen when multiple callbacks fire with error, which is likely, since once one of them does the listener will fire all the other ones too. Instead, we should probably add a flag (guarded by the mutex) telling us whether we've already set the future.

tensorpipe/test/channel/channel_test.h
```diff
@@ -203,39 +203,51 @@ class ChannelTestCase {
...
 class ClientServerChannelTestCase : public ChannelTestCase<TBuffer> {
-  std::vector<std::future<std::shared_ptr<tensorpipe::transport::Connection>>>
-  accept(
-      std::shared_ptr<tensorpipe::transport::Listener> listener,
-      size_t numConnections) {
-    auto promises = std::make_shared<std::vector<
-        std::promise<std::shared_ptr<tensorpipe::transport::Connection>>>>(
+  std::future<std::pair<
+      tensorpipe::Error,
+      std::vector<std::shared_ptr<tensorpipe::transport::Connection>>>>
```
lw, line 208: Nit: A type spanning three lines starts feeling like something we could address. Especially when one of these types is std::pair/tuple, which is often a shortcut for what could be a custom struct. So this could become `std::future<MultiAcceptResult>`, with `MultiAcceptResult` being a struct with two fields, `error` and `connections`.

tensorpipe/test/channel/channel_test.h
```diff
@@ -203,39 +203,51 @@ class ChannelTestCase {
...
+  accept(tensorpipe::transport::Listener& listener, size_t numConnections) {
+    auto promise = std::make_shared<std::promise<std::pair<
+        tensorpipe::Error,
+        std::vector<std::shared_ptr<tensorpipe::transport::Connection>>>>>();
+    auto connections = std::make_shared<
+        std::vector<std::shared_ptr<tensorpipe::transport::Connection>>>(
         numConnections);
+    auto connectionsMutex = std::make_shared<std::mutex>();
```
lw: Nit: A similar comment applies here, we're creating three separate shared_ptrs, which we then however always pass around together. It may make sense to "bundle" these values into a single struct and allocate a single shared_ptr.

lw (approved):

tensorpipe/test/channel/channel_test.cc
```diff
@@ -356,36 +310,7 @@ class SendTensorsBothWaysTest : public ClientServerChannelTestCase<TBuffer> {
...
-        this->helper_->makeContext("client");
-    EXPECT_EQ(kReady, this->peers_->recv(PeerGroup::kClient));
-    context->createChannel({std::move(conn)}, Endpoint::kConnect);
-  }
-};
-
-template <typename TBuffer>
-const std::string ContextIsNotJoinedTest<TBuffer>::kReady = "ready";
-
-CHANNEL_TEST_GENERIC(ContextIsNotJoined);
```
lw, line 391: Sorry for changing my mind, but could we keep this around? It's the only test that triggers the TSAN error on MPT, which I find suspicious and I want one day to investigate it well.

beauby: I added the test back exactly as it was before.

lw (approved):

PR #291, opened 2021-02-04
[shm] Dynamically check if MAP_SHARED_VALIDATE supported

malfet:
TensorPipe can be compiled against newer glibc, but launched on older kernel,
where MAP_SHARED_VALIDATE is equivalent to MAP_SHARED|MAP_PRIVATE, which renders every `mmap` invalid
Check that MAP_SHARED_VALIDATE with non-existent flag against stdout returns EOPNOSUPP rather than EINVAL
Fixes issue discovred while runner Ubuntu-18.04 in Docker container, see https://app.circleci.com/pipelines/github/pytorch/pytorch/268730/workflows/6d420dc6-7795-4ef8-9639-d9e1439a1f1a/jobs/10631993

### DIFF
--- a/tensorpipe/util/shm/segment.cc
+++ b/tensorpipe/util/shm/segment.cc
@@ -62,13 +62,25 @@ constexpr PageType getDefaultPageType(uint64_t size) {
   }
 }
 
+#ifdef MAP_SHARED_VALIDATE
+bool isMapSharedValidateSupported() {
+  static bool rc = []() {
+    // If MAP_SHARED_VALIDATE is supported by kernel,
+    // errno should be EOPNOTSUPP, if not EINVAL
+    void *ptr = mmap(NULL, 8, 0, 0x800000 | MAP_SHARED_VALIDATE, 0, 0);
+    return ptr == MAP_FAILED && errno == EOPNOTSUPP;
+  }();
+  return rc;
+}
+#endif
+
 std::tuple<Error, MmappedPtr> mmapShmFd(
     int fd,
     size_t byteSize,
     bool permWrite,
     optional<PageType> pageType) {
 #ifdef MAP_SHARED_VALIDATE
-  int flags = MAP_SHARED | MAP_SHARED_VALIDATE;
+  int flags = isMapSharedValidateSupported() ? MAP_SHARED_VALIDATE : MAP_SHARED;
 #else
 
 #warning \


tensorpipe/util/shm/segment.cc
```diff
@@ -62,13 +62,25 @@ constexpr PageType getDefaultPageType(uint64_t size) {
   }
 }
 
+#ifdef MAP_SHARED_VALIDATE
+bool isMapSharedValidateSupported() {
+  static bool rc = []() {
+    // If MAP_SHARED_VALIDATE is supported by kernel,
+    // errno should be EOPNOTSUPP, if not EINVAL
+    void *ptr = mmap(NULL, 8, 0, 0x800000 | MAP_SHARED_VALIDATE, 0, 0);
+    return ptr == MAP_FAILED && errno == EOPNOTSUPP;
```
lw, line 71: I am a tad concerned that this partly relies on undefined behavior: it relies on `0x800000` not being a valid flag (and it would start returning a false negative once it is), and it assumes the flags are validated before other fields (like the fd, which might cause the error to be EACCESS or EBADF).

I'd be a bit more reassured if instead we used a fully valid call here, and instead checked for success-vs-failure (rather than between two modes of failure). For example, we could use MAP_ANONYMOUS and specify an fd of -1, which should then always succeed.
```suggestion
    void *ptr = ::mmap(/*addr=*/nullptr, /*length=*/8, PROT_NONE, MAP_SHARED_VALIDATE | MAP_ANONYMOUS, /*fd=*/-1, /*offset=*/0);
    return ptr != MAP_FAILED;
    // FIXME munmap if successful
```

lw (approved):
Thanks for the catch and the fix! This could land as-is, but I've also left a suggestion improvement in case.

tensorpipe/util/shm/segment.cc
malfet: That was my initial though, but alas, `mmap(NULL, 8, 0, MAP_ANONYMOUS | MAP_SHARED_VALIDATE, -1, 0)` returns -EINVAL on both supported an unsupported kernels (because anonymous mappings can be shared).
Another alternative would be to use `uname()` syscall and return true only if kernel version is greater than 4.14

lw: Weird, the man page for [mmap](https://man7.org/linux/man-pages/man2/mmap.2.html) says:

> The use of MAP_ANONYMOUS in conjunction with MAP_SHARED is supported on Linux only since kernel 2.4.

If you can't get this to work then never mind, you can land it as is.

PS: Maybe it's an issue with the PROT? Like, the kernel refuses to make an anonymous mapping to which the only user has neither read nor write permissions, as that's useless?

malfet: MAP_SHARED_VALIDATE would not work with anonymous maps, see 
https://github.com/torvalds/linux/blob/61556703b610a104de324e4f061dc6cf7b218b46/mm/mmap.c#L1565

PR #303, opened 2021-02-22
Optimize pinned memory allocation in CudaBasic.

beauby:
* **#303 Optimize pinned memory allocation in CudaBasic.**

This PR introduces CudaHostAllocator, which allocates a single slab of
pinned host memory. It works by providing fixed-size chunks of memory
for use by the CudaBasic channel.

The CudaBasic channel itself is being reworked to chunk large input
tensors (so that each chunk fits into a pinned memory chunk), and send
each one separately, in order, over the the underlying CPU channel.

NOTE: At the moment, the CudaHostAllocator will block a whole chunk
for each allocation, possibly leading to under-used memory (for
instance when sending several small tensors). It is possible to
mitigate this by fitting multiple allocations per chunk by maintaining
two variables per chunk: `offset` and `used`, the former being
non-decreasing. Each time an allocation for `size` bytes is requested,
each chunk is checked, and if `chunkSize-offset >= size`, `offset` and
`used` are both incremented by `size`. When an allocation is being
reclaimed, the `used` variable is decremented by `size`, and if it
falls to zero, `offset` is reset to zero as well.

Differential Revision: [D26575581](https://our.internmc.facebook.com/intern/diff/D26575581)

### DIFF
--- a/tensorpipe/channel/cuda_basic/channel_impl.cc
+++ b/tensorpipe/channel/cuda_basic/channel_impl.cc
@@ -137,7 +137,7 @@ void ChannelImpl::sendChunkThroughCpuChannel(Operation op) {
   // FIXME: Avoid copying the op twice.
   cpuChannel_->send(
       cpuBuffer,
-      callbackWrapper_([op](ChannelImpl& impl, std::string descriptor) {
+      callbackWrapper_([op](ChannelImpl& impl, std::string descriptor) mutable {
         if (impl.error_) {
           return;
         }
--- a/tensorpipe/channel/cuda_basic/channel_impl.cc
+++ b/tensorpipe/channel/cuda_basic/channel_impl.cc
@@ -156,6 +156,7 @@ void ChannelImpl::sendChunkDescriptor(Operation op, std::string descriptor) {
              << op.sequenceNumber;
   connection_->write(
       &descriptor[0],
+      descriptor.length(),
       callbackWrapper_([op{std::move(op)},
                         descriptor{std::move(descriptor)}](ChannelImpl& impl) {
         TP_VLOG(6) << "Channel " << impl.id_
--- a/tensorpipe/channel/cuda_basic/channel_impl.cc
+++ b/tensorpipe/channel/cuda_basic/channel_impl.cc
@@ -98,7 +98,7 @@ void ChannelImpl::sendImplFromLoop(
     cudaHostAllocator_.alloc(
         op.length,
         callbackWrapper_(
-            [&op](ChannelImpl& impl, std::shared_ptr<uint8_t[]> tmpBuffer) {
+            [&op](ChannelImpl& impl, std::unique_ptr<uint8_t[]> tmpBuffer) {
               TP_VLOG(5) << "Channel " << impl.id_
                          << " is done allocating temporary memory for chunk #"
                          << op.chunkId << " of " << op.numChunks
@@ -243,7 +243,7 @@ void ChannelImpl::onRecvOpReadDescriptor(
       op.length,
       callbackWrapper_(
           [&op, descriptor{std::move(descriptor)}](
-              ChannelImpl& impl, std::shared_ptr<uint8_t[]> tmpBuffer) mutable {
+              ChannelImpl& impl, std::unique_ptr<uint8_t[]> tmpBuffer) mutable {
             TP_VLOG(5) << "Channel " << impl.id_
                        << " is done allocating temporary memory for chunk #"
                        << op.chunkId << " of " << op.numChunks
--- a/tensorpipe/channel/cuda_basic/channel_impl.cc
+++ b/tensorpipe/channel/cuda_basic/channel_impl.cc
@@ -295,9 +295,10 @@ void ChannelImpl::onRecvOpReadyForCopy(Operation& op) {
         TP_VLOG(5) << "Channel " << impl.id_ << " is done copying chunk #"
                    << op.chunkId << " of " << op.numChunks << " for buffer #"
                    << op.sequenceNumber << " from CPU to CUDA device";
-        op.done = true;
-        impl.onRecvOpDone();
       }));
+
+  op.done = true;
+  impl.onRecvOpDone();
 }
 
 void ChannelImpl::onRecvOpDone() {
--- a/tensorpipe/channel/cuda_basic/channel_impl.cc
+++ b/tensorpipe/channel/cuda_basic/channel_impl.cc
@@ -291,7 +291,8 @@ void ChannelImpl::onRecvOpReadyForCopy(Operation& op) {
       op.length,
       deviceIdx,
       op.stream,
-      callbackWrapper_([&op](ChannelImpl& impl) {
+      // TODO: Avoid copying op.
+      callbackWrapper_([op](ChannelImpl& impl) {
         TP_VLOG(5) << "Channel " << impl.id_ << " is done copying chunk #"
                    << op.chunkId << " of " << op.numChunks << " for buffer #"
                    << op.sequenceNumber << " from CPU to CUDA device";
--- a/tensorpipe/channel/cuda_basic/channel_impl.cc
+++ b/tensorpipe/channel/cuda_basic/channel_impl.cc
@@ -19,6 +19,7 @@
 #include <tensorpipe/common/cuda.h>
 #include <tensorpipe/common/defs.h>
 #include <tensorpipe/common/error.h>
+#include <tensorpipe/transport/connection.h>
 
 namespace tensorpipe {
 namespace channel {
@@ -30,17 +31,34 @@ ChannelImpl::ChannelImpl(
     std::string id,
     std::shared_ptr<transport::Connection> connection,
     std::shared_ptr<CpuChannel> cpuChannel,
-    CudaLoop& cudaLoop)
+    CudaLoop& cudaLoop,
+    CudaHostAllocator& cudaHostAllocator)
     : ChannelImplBoilerplate<CudaBuffer, ContextImpl, ChannelImpl>(
           token,
           std::move(context),
           std::move(id)),
       connection_(std::move(connection)),
       cpuChannel_(std::move(cpuChannel)),
-      cudaLoop_(cudaLoop) {}
+      cudaLoop_(cudaLoop),
+      cudaHostAllocator_(cudaHostAllocator) {}
 
 void ChannelImpl::initImplFromLoop() {
   context_->enroll(*this);
+}
+
+void ChannelImpl::cudaCopy(
+    void* dst,
+    const void* src,
+    size_t length,
+    int deviceIdx,
+    cudaStream_t stream,
+    std::function<void(const Error&)> callback) {
+  {
+    CudaDeviceGuard guard(deviceIdx);
+    TP_CUDA_CHECK(cudaMemcpyAsync(dst, src, length, cudaMemcpyDefault, stream));
+  }
+
+  cudaLoop_.addCallback(deviceIdx, stream, std::move(callback));
 }
 
 void ChannelImpl::sendImplFromLoop(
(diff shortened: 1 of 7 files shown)


tensorpipe/channel/cuda_basic/channel_impl.cc
```diff
@@ -30,165 +31,319 @@ ChannelImpl::ChannelImpl(
...
-      TP_VLOG(6) << "Channel " << id_ << " is sending buffer #"
-                 << op.sequenceNumber << " through CPU channel";
-      cpuChannel_->send(
-          cpuBuffer, std::move(op.descriptorCallback), std::move(callback));
+    if (op.callback) {
+      op.callback(error_);
+    }
+    if (!error_) {
+      sendChunkThroughCpuChannel(std::move(op));
     }
```
beauby, line 180: One thing I'm not sure is what happens if the context is being closed once the send callback fires, but before the tensor is actually sent through the CPU channel?

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -0,0 +1,120 @@
...
+
+CudaHostAllocator::CudaHostAllocator(size_t numChunks, size_t chunkSize)
+    : numChunks_(numChunks),
+      chunkSize_(chunkSize),
+      data_(allocPinnedBuffer(numChunks * chunkSize)),
+      chunkAvailable_(numChunks, true) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  thread_ = std::thread([this, lock{std::move(lock)}]() mutable {
+    setThreadName("TP_CUDA_host_allocator_loop");
+    allocLoop(std::move(lock));
```
lw, line 25: You had used this pattern before (acquire lock in constructor, move into main loop thread), and I think we concluded that it wasn't necessary. I've got the same feeling now, as you sure this is needed?

The reason for my feeling is that the main thread will inevitably end up releasing and then re-acquiring the lock at some point, typically during wait, and thus needs to be able to determine on its own whether it is able to perform some work or not (in particular, it needs to do so before deciding to go to sleep). Hence that thread should be able to cope with the fact of starting up without already holding the lock.

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -0,0 +1,120 @@
...
+    thread_.join();
+  }
+}
+
+void CudaHostAllocator::allocLoop(std::unique_lock<std::mutex> lock) {
+  for (;;) {
+    if (closed_ && pendingAllocations_.empty()) {
+      break;
+    } else {
+      cv_.wait(lock);
```
lw, line 69: This wake-up can happen in three cases: new request from the user, a previous chunk was returned to the pool, and the allocator was closed. The last case (i.e., closed) seems it could deserve a dedicated check after this wakeup?

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -0,0 +1,120 @@
...
+    processAllocations(allocations);
+    lock.lock();
+    pendingAllocations_.insert(
+        pendingAllocations_.begin(), allocations.begin(), allocations.end());
+  }
+}
+
+void CudaHostAllocator::processAllocations(
+    std::deque<TAllocCallback>& allocations) {
+  for (size_t curChunk = 0; curChunk < numChunks_; ++curChunk) {
```
lw, line 84: Is this the most efficient way of going about this? As it stands, we'd risk going over all chunks (a 1000 of them, according to the default values) if the one available chunk is the last one.

We don't really need to know the state of each chunk, all we need to know is which chunks are available. What if we kept a `std::deque<int>` that stored the _indices_ of the available chunks? In `hostPtrDeleter` we would push a new index to the back of that deque, whereas in this method we'd pop from the front until non-empty.

A deque might add some overhead, as it causes allocs and frees when pushing and popping. As in our case the deque size is bounded and known in advance (i.e., up to the total number of chunks), we could reimplement it more efficiently, e.g., as a vector-backed ringbuffer (i.e., push becomes `freeList_[(head++) % numChunks_] = value`, pop becomes `freeList_[(tail++) % numChunks_] if tail < head else -1`).

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -0,0 +1,120 @@
...
+size_t CudaHostAllocator::getChunkLength() const {
+  return chunkSize_;
+}
+
+void CudaHostAllocator::close() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  if (closed_) {
+    return;
+  }
+  closed_ = true;
```
lw, line 53: I think there's something odd going on here. After closing, all new requests are immediately called with error (and that's good), however what happens to the existing ones? We wake up the loop, which will do one additional iteration, possibly serving some of them. However at the next iteration the loop will return, leaving all pending requests unserved, and then never touching them again. This seems incorrect...

lw (commented):
I've only looked at the allocator itself, not yet at how it's used in the channel, hence I'm sending out a partial review.

Some random comments (more food for thoughts than requests for chanes):

It's an interesting choice to implement the "return to pool" step implicitly in the shared_ptr's destructor. I like that it acts as a safety net, making it impossible to thus "leak" memory. However I'd also prefer if it was more explicit, especially if/when we'll use a state machine in CUDA basic, as that will allow us to express it as a transaction in the state machine and clarify the moment it's supposed to happen. Ideally it would be cool if we could get the best of both worlds, and perhaps we can, by "explicitly" resetting the shared_ptr to nullptr when we want to return the memory to the pool.

I'm still not really convinced we need a thread to power the allocator (we already use quite a bunch of threads, and any one we can avoid is probably a good thing). All calls to the public methods will in practice happen from the same deferred executor, hence there's no risk for concurrency. And any "event" (like memory being requested or returned to the pool) is triggered by one of these public methods, and thus we could just handle that event "inline" rather than deferring to the thread.

After suggesting to use a deque for the free-list of chunks, I realized that the design of this allocator resembles a lot the RearmableCallback helper we have in callback.h. Both of them rely on the same principle: we have a set of available resources we're giving out (in this case, the chunks) and a list of consumers who have requested them. The invariant is that every point in time at least one of these two sets must be empty: if it isn't, then it means we can "satisfy" one of the consumer requests by passing to it one of the available resources (and pop from both sets). I wonder if the RearmableCallback could be used here to simplify the code.

tensorpipe/channel/cuda_basic/channel_impl.cc
```diff
@@ -30,165 +31,319 @@ ChannelImpl::ChannelImpl(
...
+    // Operations are processed in order, so we can afford to trigger the
+    // callback once the last operation is done.
+    if (op.chunkId == numChunks - 1) {
+      op.callback = std::move(callback);
+    }
+
+    TP_VLOG(6) << "Channel " << id_ << " is reading descriptor for chunk #"
+               << op.chunkId << " of " << op.numChunks << " for buffer #"
+               << op.sequenceNumber;
+    auto nopHolderIn = std::make_shared<NopHolder<std::string>>();
```
lw, line 214: Same here

tensorpipe/channel/cuda_basic/channel_impl.cc
```diff
@@ -30,165 +31,319 @@ ChannelImpl::ChannelImpl(
...
+                     << op.sequenceNumber;
+          std::string& descriptor = nopHolderIn->getObject();
+          impl.onRecvOpReadDescriptor(op, std::move(descriptor));
+        }));
+  }
+}
+
+void ChannelImpl::onRecvOpReadDescriptor(
+    Operation& op,
+    std::string descriptor) {
```
lw, line 230: The fact that here (and several more times below) we're always passing the descriptor along with the operation really gives a feeling that the descriptor should just be a field of the operation.

tensorpipe/channel/cuda_basic/channel_impl.cc
```diff
@@ -30,165 +31,319 @@ ChannelImpl::ChannelImpl(
...
 
 void ChannelImpl::setIdImpl() {
   cpuChannel_->setId(id_ + ".cpu");
 }
 
+void ChannelImpl::tryCleanup() {
+  TP_DCHECK(context_->inLoop());
+  TP_DCHECK(error_);
+
+  if (sendOperations_.empty() && recvOperations_.empty()) {
```
lw, line 326: I'm  bit puzzled by this, could you explain why it's needed?

lw (commented):
I find the logic of the channel to have become very convoluted and hard to follow, with a lot of branching and jumps and "merges". For example, I am far from sure that everything proceeds as intended in case of error, as errors are handled in several place, as operations remain alive (by being moved around) even after they're removed from the deque, etc.

I think the sane solution to this is a state machine (unsurprisingly). The question is whether we want to rework this PR to already use the state machine, or whether we want to make a "leap of faith" and land this as-is (hoping there's no bugs) and convert it to the state machine immediately after.

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -51,70 +55,62 @@ void CudaHostAllocator::close() {
...
 
-void CudaHostAllocator::allocLoop(std::unique_lock<std::mutex> lock) {
-  for (;;) {
-    if (closed_ && pendingAllocations_.empty()) {
-      break;
+void CudaHostAllocator::processAllocations(std::unique_lock<std::mutex> lock) {
+  while (!pendingAllocations_.empty()) {
+    auto& callback = pendingAllocations_.front();
+    if (closed_) {
+      callback(TP_CREATE_ERROR(CudaHostAllocatorClosedError), nullptr);
```
lw, line 75: Shouldn't we release the lock while we invoke the callback? This has bitten us in the past...

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -51,70 +55,62 @@ void CudaHostAllocator::close() {
...
 }
 
 void CudaHostAllocator::join() {
   close();
   if (!joined_.exchange(true)) {
-    thread_.join();
+    std::unique_lock<std::mutex> lock(mutex_);
+    cv_.wait(lock, [this]() {
+      return pendingAllocations_.empty() && (allocatedChunks_ == 0);
+    });
```
lw, line 67: What is the point of this join method? It waits without doing anything

lw (commented):
Let me start with some comments on the allocator (will look at the channel in a sec). While I'm happy we removed the thread, I'm surprised we weren't able to remove more of the "related" stuff, like the mutex and the condition variable. I think they do introduce risks (e.g., not releasing it when invoking a callback) and they are only needed to synchronize the join method with the rest of the class (hence they are never needed during "normal" operations). If we released the pinned memory in the destructor (which, actually, is what we're already doing, probably by accident) then we wouldn't need the join method at all and we could simplify things further.

tensorpipe/channel/cuda_basic/channel_impl.cc
```diff
@@ -118,67 +195,124 @@ void ChannelImpl::recvImplFromLoop(
...
-                   << sequenceNumber << " from CPU to CUDA device";
+  int deviceIdx = cudaDeviceForPointer(context_->getCudaLib(), op.cudaPtr);
+  cudaCopy(
+      op.cudaPtr,
+      op.tmpBuffer.get(),
+      op.length,
+      deviceIdx,
+      op.stream,
+      // TODO: Avoid copying op.
+      callbackWrapper_([op](ChannelImpl& impl) {
```
lw, line 295: Correct me if I'm wrong, but this callback is performing the crucial duty of keeping the operation alive (and thus delaying when the pinned staging buffer is returned to the allocator), right? This is far from obvious by looking at this in isolation, and based on that TODO one would be tempted to "fix" this code by avoiding to capture the op and instead capture the `chunkId`, `numChunks` and `sequenceNumber` fields separately, as these appear to be the only ones that are needed.

I think we'll fix this "automatically" with the state machine, but in the meantime it would be nice to at least add a comment about this to prevent accidental modifications.

lw (approved):

tensorpipe/common/cuda_host_allocator.cc
```diff
@@ -14,31 +14,35 @@
...
-}
+      data_(allocPinnedBuffer(numChunks * chunkSize), freePinnedBuffer),
+      chunkAvailable_(numChunks, true) {}
 
 CudaHostAllocator::~CudaHostAllocator() {
   join();
 }
 
 void CudaHostAllocator::alloc(size_t size, TAllocCallback callback) {
   std::unique_lock<std::mutex> lock(mutex_);
```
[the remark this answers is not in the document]
beauby: Right, so as long as `alloc()`, `close()`, and `releaseChunk()` are always called from the context loop, we can get rid of the mutex. Removing.

tensorpipe/common/cuda_host_allocator.cc
beauby: The point was to ensure the Allocator remains alive until all chunks have been reclaimed, but we can rely on the Context outliving the channels for this. Removing.

tensorpipe/channel/cuda_basic/channel_impl.cc
beauby: Added comment.

tensorpipe/common/cuda_host_allocator.cc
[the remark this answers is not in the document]
lw: If you want to make this assumption explicit, hence make it both safer and clearer to the future readers of this code, you could make it so that the constructor of the allocator takes a const ref to a DeferredExecutor, store it, and here call `TP_DCHECK(loop_.inLoop())`.

tensorpipe/common/cuda_host_allocator.cc
beauby: Got rid of the lock altogether.

PR #305, opened 2021-02-23
Add MiscCheck.cmake

walterddr:
Usage of `::process_vm_readv` is added in 622bdb9a3b1b02b53e988377118e36af2f391088 .
However the function is not available in `<sys/uio.h>` for some older libc distributions such as glibc 2.12. 

This PR 
- Refactors and adds cmake/MiscCheck.cmake file for compilation checker. 
- Adds a checker to explicitly catch libc version error.

Test Plan
CI

### DIFF
--- a/cmake/MiscCheck.cmake
+++ b/cmake/MiscCheck.cmake
@@ -9,39 +9,41 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   endif()
 endif()
 
-# Check libc contains process_vm_readv
-CMAKE_PUSH_CHECK_STATE(RESET)
-set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS}")
-CHECK_CXX_SOURCE_COMPILES("
-#include <sys/uio.h>
-#include <sys/types.h>
-#include <unistd.h>
+if(LINUX)
+  # Check libc contains process_vm_readv
+  CMAKE_PUSH_CHECK_STATE(RESET)
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS}")
+  CHECK_CXX_SOURCE_COMPILES("
+    #include <sys/uio.h>
+    #include <sys/types.h>
+    #include <unistd.h>
 
-namespace {
-bool isProcessVmReadvSyscallAllowed() {
-  long long someSourceValue = 0x0123456789abcdef;
-  long long someTargetValue = 0;
-  struct iovec source = {
-    .iov_base = &someSourceValue, .iov_len = sizeof(someSourceValue)
-  };
-  struct iovec target = {
-    .iov_base = &someTargetValue, .iov_len = sizeof(someTargetValue)
-  };
-  ssize_t nread = ::process_vm_readv(::getpid(), &target, 1, &source, 1, 0);
-  return nread == sizeof(long long) && someTargetValue == someSourceValue;
-}
-}
+    namespace {
+    bool isProcessVmReadvSyscallAllowed() {
+      long long someSourceValue = 0x0123456789abcdef;
+      long long someTargetValue = 0;
+      struct iovec source = {
+        .iov_base = &someSourceValue, .iov_len = sizeof(someSourceValue)
+      };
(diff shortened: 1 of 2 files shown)


lw:
Thanks for the PR! I'd like to understand where this issue occurs. The glibc version you mentioned, 2.12, seems to date from 2010, hence it's almost 11 years old. As a rule of thumb, we're supporting dependencies up to the version they had in Ubuntu 16.04 (5 years ago), and there glibc was at version 2.27. For example, we require C++-11 to build, which didn't even exist yet back in 2010. Is there a specific reason you're finding yourself with such old glibc version? Perhaps a specific use case I didn't consider? If not, I'd be inclined to not consider out-of-scope the support for such old versions.

walterddr:
this is for one of the centos6 toolchains that just got deprecated last Nov. 2020. However, conda gxx_linux-64 is still depending on this.

lw:
Sorry, I'm not familiar with conda. It does indeed seem that CentOS 6 had glibc 2.12, but it also had GCC 4.4, which we don't support either. How come conda uses only "half" of that toolchain?

Another question is how does this PR address that issue? It seems to me that it just replaces a compile-time error (the missing `process_vm_readv` declaration) with a CMake-time fatal error. So is the goal of this just to provide an earlier failure?

walterddr:
thanks for the information. @lw.
- For [conda toolchain](https://anaconda.org/conda-forge/gxx_linux-64) it is actually installing GCC 7-9 based of centos6. 
- Regarding the fix, I didn't put it in the PR b/c I am not super familiar with TP repo. but I was thinking maybe we can turn off `TP_ENABLE_CMA` if glibc version is too old? however I am not sure if that's even an acceptable solution on linux. if you could share your thoughts on this that would be super great.

malfet (approved):

cmake/MiscCheck.cmake
```diff
@@ -0,0 +1,50 @@
...
+      isProcessVmReadvSyscallAllowed();
+      return 0;
+    }" SUPPORT_GLIBCXX_USE_PROCESS_VM_READV)
+  if(NOT SUPPORT_GLIBCXX_USE_PROCESS_VM_READV)
+    unset(SUPPORT_GLIBCXX_USE_PROCESS_VM_READV CACHE)
+    message(WARNING
+        "The C++ compiler does not support required functions. "
+        "This likely due to an old version of libc.so is used. "
+        "Please check your system linker setting.")
+    set(TP_ENABLE_CMA OFF)
```
lw, line 46: The approach we've taken with other backends is to auto-detect their availability and turn them off by default if any issues are encountered, but to error out if the user explicitly requested for them to be built, as this would silently give users a worse behavior than they requested. I don't think printing a warning would be enough: no one really reads compilation logs.

cmake/MiscCheck.cmake
```diff
@@ -0,0 +1,50 @@
...
+      struct iovec source = {
+        .iov_base = &someSourceValue, .iov_len = sizeof(someSourceValue)
+      };
+      struct iovec target = {
+        .iov_base = &someTargetValue, .iov_len = sizeof(someTargetValue)
+      };
+      ssize_t nread = ::process_vm_readv(::getpid(), &target, 1, &source, 1, 0);
+      return nread == sizeof(long long) && someTargetValue == someSourceValue;
+    }
+    }
```
lw, line 34: This code is just required to compile right? It doesn't have to be be correct when executed? In that case, we can probably simplify this simply to:
```
int main() {
  ssize_t rv = ::process_vm_readv(1, nullptr, 0, nullptr, 0, 0);
  return 0;
}
```
Plus possibly casting arguments to the right types for extra safety.

lw (commented):
Ok, if support for such old versions of glibc is required by PyTorch let's go for it. Probably the best behavior here is to disable the CMA backend when that syscall is unavailable. This should be fine (it's for example what we already do on OSX).

cmake/MiscCheck.cmake
walterddr: yes. that's a good point. we should simplify, sorry for just copy-pasting the code from context_impl.cc

PR #312, opened 2021-03-10
Merge channel hierarchies.

beauby:
* **#312 Merge channel hierarchies.**

This is a first step towards cross-device type transfers: eventually,
channels will not connect devices of a given type between two hosts,
but possibly heterogeneous pairs of devices. Hence, the distinction
between CPU-to-CPU and GPU-to-GPU channels will not make much sense
anymore, and we can afford to simplify the Pipe's code quite bit.

The main change here is that the `channel::Channel` and
`channel::Context` classes are not templated (on the buffer type)
anymore. Instead, a channel's `send`/`recv` methods act on generic
`Buffer`s and the actual unpacking is done in the
`ChannelBoilerplate`. The
`channel::CpuContext`/`channel::CudaContext` (respectively
`channel::CudaContext`/`channel::CudaChannel`) aliases now simply
resolve to `channel::Context` (respectively `channel::Channel`). A
subsequent diff will get rid of the aliases altogether.

The Pipe is being simplified: all the duplication due to having
separate hierarchies is gone, which gets rid of a lot of boiler plate
template code. Note that previously, two channels with the same name
could potentially coexist, provided one was a CPU channel and the
other a GPU channel. This is not the case anymore, though it should
not matter.
In its current state, the Pipe still needs to pick a channel based on
whether that channel acts on CPU or GPU buffers. This is solved by
introducing the temporary method
`bool channel::Context::supportsDeviceType(DeviceType t)`. When
iterating through available channels to select one for a given tensor,
the Pipe now discards channels that do not support the tensor's
`DeviceType`. This leads to having a single ordered list of channels,
which in practice is two separate lists (one for CPU, one for GPU)
merged together. This will change soon as we initialize only one
channel per `DeviceType`.

Differential Revision: [D26958187](https://our.internmc.facebook.com/intern/diff/D26958187)


tensorpipe/core/buffer.h
```diff
@@ -59,4 +60,21 @@ struct Buffer {
   };
 };
 
+template <typename TBuffer>
+TBuffer unwrap(Buffer);
```
beauby, line 64: This is now used not only by the `PipeImpl` but also by the `ChannelBoilerplate`, hence it needs to be declared in a header.

tensorpipe/channel/channel_boilerplate.h
```diff
@@ -82,18 +82,22 @@ ChannelBoilerplate<TBuffer, TCtx, TChan>::ChannelBoilerplate(
 
 template <typename TBuffer, typename TCtx, typename TChan>
 void ChannelBoilerplate<TBuffer, TCtx, TChan>::send(
-    TBuffer buffer,
+    Buffer buffer,
     TDescriptorCallback descriptorCallback,
     TSendCallback callback) {
-  impl_->send(buffer, std::move(descriptorCallback), std::move(callback));
+  impl_->send(
+      unwrap<TBuffer>(buffer),
```
beauby, line 89: Buffer unwrapping now happens here.

tensorpipe/channel/cpu_context.h
```diff
@@ -9,13 +9,14 @@
...
 
 namespace tensorpipe {
 namespace channel {
 
-using CpuChannel = Channel<CpuBuffer>;
-using CpuContext = Context<CpuBuffer>;
+// FIXME: These aliases are temporarily required for backwards compatibility.
+// Remove once dependent code has been updated.
+using CpuChannel = Channel;
+using CpuContext = Context;
```
beauby, line 19: These aliases are kept for now so that 1. the public API remains backwards compatible, and 2. the line count for this PR is kept low.

tensorpipe/core/pipe_impl.cc
```diff
@@ -871,42 +750,48 @@ void PipeImpl::sendTensorsOfMessage(WriteOpIter opIter) {
...
-      return WriteOperation::Tensor{};
-    });
-    op.tensors.push_back(t);
+      auto& channelContext = std::get<1>(channelContextIter.second);
+      // FIXME: Skip channel if it does not support current buffer type, as a
+      // temporary measure before we roll out proper channel selection for
+      // cross-device type transfers.
+      if (!channelContext->supportsDeviceType(tensor.buffer.type)) {
+        continue;
+      }
```
beauby, line 767: This is where we skip channels based on supported `DeviceType`.

tensorpipe/channel/context.h
```diff
@@ -97,6 +96,10 @@ class Context {
 
   virtual ~Context() = default;
 
+  // FIXME: This is a private, temporary, API. It will be removed in the
+  // coming weeks. DO NOT USE IT.
+  virtual bool supportsDeviceType(DeviceType type) const = 0;
```
beauby, line 101: Here is the temporary method for determining which type of buffer a channel supports.

lw (approved):

beauby:
Closing in favor of #326.

Issue #316, opened 2021-03-14
Build fails with clang and libc++ on Linux


SpaceIm:
Several includes are missing in `tensorpipe/util/ringbuffer/consumer.h` and `tensorpipe/util/ringbuffer/producer.h` leading to these errors:

<details><summary>Click to expand log</summary>

```
 -- The C compiler identification is Clang 9.0.0
-- The CXX compiler identification is Clang 9.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Conan: called by CMake conan helper
-- Conan: called inside local cache
-- Conan: Adjusting output directories
-- Conan: Using cmake global configuration
-- Conan: Adjusting default RPATHs Conan policies
-- Conan: Adjusting language standard
-- Conan: Adjusting fPIC flag (ON)
-- Conan: Compiler Clang>=8, checking major version 9
-- Conan: Checking correct version: 9
-- Conan: C++ stdlib: libc++
-- Performing Test SUPPORT_GLIBCXX_USE_PROCESS_VM_READV
-- Performing Test SUPPORT_GLIBCXX_USE_PROCESS_VM_READV - Success
-- Found uv: 1.41.0 (found version "1.41.0") 
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_EXPORT_NO_PACKAGE_REGISTRY


-- Build files have been written to: /home/conan/.conan/data/tensorpipe/cci.20210309/bincrafters/testing/build/7166586d157fc254fc999586d73a9400ec890d34
Scanning dependencies of target tensorpipe
[  1%] Building CXX object source_subfolder/tensorpipe/CMakeFiles/tensorpipe.dir/channel/error.cc.o
[  3%] Building CXX object source_subfolder/tensorpipe/CMakeFiles/tensorpipe.dir/channel/helpers.cc.o
[  5%] Building CXX object source_subfolder/tensorpipe/CMakeFiles/tensorpipe.dir/common/address.cc.o
[...]

PR #318, opened 2021-03-15
Add missing includes in ringbuffer headers

SpaceIm:
closes https://github.com/pytorch/tensorpipe/issues/316

### DIFF
--- a/tensorpipe/util/ringbuffer/producer.h
+++ b/tensorpipe/util/ringbuffer/producer.h
@@ -8,6 +8,12 @@
 
 #pragma once
 
+#include <algorithm>
+#include <array>
+#include <cstring>
+#include <tuple>
+#include <utility>
+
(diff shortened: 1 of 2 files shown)


tensorpipe/util/ringbuffer/producer.h
```diff
@@ -10,6 +10,12 @@
 
 #include <tensorpipe/util/ringbuffer/ringbuffer.h>
 
+#include <algorithm>
+#include <array>
+#include <cstring>
+#include <tuple>
+#include <utility>
```
lw, line 17: Just one minor nit: could you put the stdlib headers in a block before the first-party headers?

The style guide we're following has these blocks:
- [for .cc files only] the corresponding .h file
- C standard headers
- C++ stdlib headers
- third-party headers
- first-party headers

Thanks!

lw (approved):
Thanks, I'm merging this, it should land shortly!

Issue #317, opened 2021-03-14
CMake: public headers are not installed


SpaceIm:
public headers installation is missing in install target

lw:
Yeah we were already tracking this in #104 (which, as you can see, is rather old). We're all in favor of doing this, but so far it hasn't proven to be critical, as PyTorch (the main user of TensorPipe) managed to work anyways. PyTorch also restricts which versions of CMake we can support (I believe we must be compatible with CMake 3.5). TBH, we probably won't get to it shortly, but if you're willing to contribute a pull request for it it would be very much appreciated!

SpaceIm:
I see. https://github.com/pytorch/tensorpipe/issues/104 is slighlty different.

I can submit a PR when I have time.
Since I'm currently packaging libtorch and all its dependencies in conan package manager, including tensorpipe, I prefer that underlying build system of each lib was responsible of installation, so I'll have to fix this at some point.

For the moment, for tensorpipe we copy all headers of these folders:
- `tensorpipe`
- `tensorpipe/channel`,
- `tensorpipe/common`
- `tensorpipe/core`
- `tensorpipe/transport`
- `tensorpipe/util`

lw:
Yes, indeed #104 is a bit different, but I think the solution is related: currently CMake isn't aware of any headers, hence they can only be fetched from the source directory. We'd need to start tracking headers in CMake, and if we do so we can very easily specify which ones must be copied to the install directory.

We are already required to do so in our internal build system, and there we distinguish between private and public headers. (Private means it's only needed to build our own .cc files, public means it must be installed as it's transitively included by `tensorpipe/tensorpipe.h`). Based off that, I compiled a list of our current public headers:
- `channel/context.h`
- `channel/cpu_context.h`
- `channel/error.h`
- `channel/cuda_context.h` (only when building with CUDA)
- `channel/*/error.h`
- `channel/*/factory.h`
- `common/cpu_buffer.h`
- `common/error.h`
- `common/cuda_buffer.h` (only when building with CUDA)
- `core/buffer.h`
- `core/context.h`
- `core/error.h`
- `core/listener.h`
- `core/message.h`
- `core/pipe.h`
- `transport/context.h`
- `transport/error.h`
- `transport/*/error.h`
- `transport/*/factory.h`
- `transport/*/utility.h`

If you (or anyone else) plans to add this information to our CMake, it would be good to keep this distinction.

SpaceIm:
Sure it's always better to be clear on public/private headers

PR #319, opened 2021-03-15
cmake: install public headers

SpaceIm:
closes https://github.com/pytorch/tensorpipe/issues/317

I understood from https://github.com/pytorch/tensorpipe/issues/317#issuecomment-799329872 that there is more generally a need to track all headers, but it's already a good improvement to track public headers.


lw (commented):
Awesome! Thanks so much for taking care of this!

I only have one request: by building the list of headers at the very end, we're replicating some of the "logic" that we already have in place above (I'm talking about the `if(...)` and such). By having that occur twice, far apart, we have the risk of forgetting to update one of them after we update the other. I think it would be safer to build the list of headers as we go, in order to have a unique logic that builds both the source files and the headers. Therefore, could you move the new code to the appropriate locations (the core list of headers after the core list of source files, the headers for channel X after the sources for channel X, and so on...)? Thanks!

SpaceIm:
something like this?

cmake/Options.cmake
```diff
@@ -48,3 +48,5 @@ option(TP_BUILD_LIBUV "Build libuv from source" OFF)
 include(GNUInstallDirs)
 set(TP_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR} CACHE STRING "Directory in which to install libraries")
 mark_as_advanced(TP_INSTALL_LIBDIR)
+set(TP_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_INCLUDEDIR} CACHE STRING "Directory in which to install public headers")
+mark_as_advanced(TP_INSTALL_INCLUDEDIR)
```
SpaceIm, line 52: Is it something you want to keep? It's quite unusual.

lw: I must admit I don't remember why we have it. Is there harm in keeping it? What's the "preferred" way?

lw (approved):
This looks great! Thanks again!

Let me know if I can go ahead and merge it or if you want to iterate more on it.

cmake/Options.cmake
SpaceIm: It doesn't harm I think. If consumer wants to change the standard install layout (bin, lib, include, share folders), I guess it's useful, but uncommon.

SpaceIm:
@lw, no more iteration for me, you can merge if you want ;)

PR #335, opened 2021-03-28
Remove descriptor/descriptorCallback from Channel interface.

beauby:
* **#335 Remove descriptor/descriptorCallback from Channel interface.**
* #340 Avoid collecting channel descriptors in channel tests.
* #339 Avoid collecting channel descriptors in the Pipe.
* #336 In-channel descriptor exchanges for CudaBasic.
* #334 In-channel descriptor exchanges for CudaGdr.
* #333 In-channel descriptor exchanges for CudaXth.
* #332 In-channel descriptor exchanges for Xth.
* #331 In-channel descriptor exchanges for Cma.

Differential Revision: [D27434726](https://our.internmc.facebook.com/intern/diff/D27434726)


tensorpipe/core/pipe_impl.h
```diff
@@ -84,13 +83,12 @@ struct WriteOperation {
   // Progress indicators.
   enum State {
     UNINITIALIZED,
-    SENDING_TENSORS_AND_COLLECTING_DESCRIPTORS,
+    SENDING_TENSORS,
     WRITING_PAYLOADS_AND_SENDING_TENSORS,
```
lw, line 87: Can't these two states be merged now?

lw (commented):
I didn't review this fully, but can I suggest to split this into two parts? First change the channel API, while keeping the changes to the pipe (and the tests?) minimal, and then fix them separately.

A way to do so could be to have the pipe maintain the same state machine, but manually set `op.tensors[tensorIdx].descriptor = ""` after calling send (and avoid increasing `op.numTensorDescriptorsBeingCollected`).

PR #348, opened 2021-04-05
[ROCm] Changes to do Hipify to support ROCm build

pruthvistony:


lw:
Hi, thanks for the contribution, and sorry for not getting back to you earlier. We weren't expecting such a PR, and before moving ahead we'd like to understand your ideas for ROCm support. Do you expect that for ROCm we can just use the automatically-transpiled ("hipified") CUDA channels? Or do we need separate specialized channels, in case there's some functional differences? In general, TensorPipe is designed to be quite device-agnostic, hence it should be feasible to have both CUDA and ROCm channels coexist in the same codebase.

We're also wondering about long-term maintenance of ROCm support. We don't have access to ROCm machines for development nor testing, thus we'll have a hard time ensuring the ROCm build remains healthy. We're also a much smaller team than PyTorch hence we can't invest as much in ROCm-specific CI and debugging its specific issues. Is this something you plan to assist with?

pruthvistony:
> Hi, thanks for the contribution, and sorry for not getting back to you earlier. We weren't expecting such a PR, and before moving ahead we'd like to understand your ideas for ROCm support. Do you expect that for ROCm we can just use the automatically-transpiled ("hipified") CUDA channels? Or do we need separate specialized channels, in case there's some functional differences? In general, TensorPipe is designed to be quite device-agnostic, hence it should be feasible to have both CUDA and ROCm channels coexist in the same codebase.
> 
> We're also wondering about long-term maintenance of ROCm support. We don't have access to ROCm machines for development nor testing, thus we'll have a hard time ensuring the ROCm build remains healthy. We're also a much smaller team than PyTorch hence we can't invest as much in ROCm-specific CI and debugging its specific issues. Is this something you plan to assist with?

Hi,
Thanks for your comments.
The intention here is to have tensorpipe with full ROCm support. Hipification is the first step to handle the generic Cuda API to HIP API conversions, but considering the channel support for ROCm, it requires more changes towards supporting NVML api(rocm-smi APIs) as per my level of analysis. However I do agree Tensoripipe is designed well and CUDA and ROCm channels can coexist.

I belong to the ROCm pytorch framework team and will work on ROCm support for tensorpipe, I will need your support to get this working. Regarding the CI support I cant comment anything now, will check on this and get back. Hope to collaborate and make pytorch better.

lw:
Thanks for the context! The thinking behind my question was that, since the orchestration logic of TensorPipe has no dependency on CUDA (only our channels do), if we'll end up having separate ROCm channels we might not need to HIPify anything at all. (I haven't work with HIPified code before and my instinct is to be a bit wary about adding this extra level of code transformation).

pruthvistony:
An updated patch is raised in PR - https://github.com/pytorch/tensorpipe/pull/367
Closing this PR for now.
Checking on the approach to support ROCm without hipify.

PR #352, opened 2021-04-12
Avoid storing length of payload/tensor in ReadOperation::Tensor.

beauby:
Summary:
This information is redundant with that of `op.message`. Moreover, once
we get rid of the need to store the `channelName` in the
`ReadOperation`/`WriteOperation`, we can fully rely on `op.message` for
information about payloads/tensors.

Reviewed By: lw

Differential Revision: D27648004


beauby:
Merged.

PR #353, opened 2021-04-12
Introduce new Pipe::{read,write} API.

beauby:
Summary:
In this diff we introduce the new API where `read()`/`write()` take a
callback that has only the error as argument (rather than the error and the
`Message`). This will ease the transition to the new API.

Reviewed By: lw

Differential Revision: D27650834


beauby:
Merged.

PR #355, opened 2021-04-12
Prepare benchmarks/benchmark_pipe.cc to new Pipe API.

beauby:
Reviewed By: lw

Differential Revision: D27650064


beauby:
Merged.

PR #357, opened 2021-04-12
Introduce Message::Tensor::targetDevice.

beauby:
Summary:
This field will remain in Message and Descriptor. It is meant for the
user to specify a target device on which a given tensor must be received.

Differential Revision: D27691722


beauby:
Merged.

PR #371, opened 2021-04-26
Add misc tool to dump state machines in dot format.

beauby:
This diff adds an executable in `misc/dump_state_machine` that takes
as parameter a `.cc` file containing the definition of a state
machine's transition table, as well as the the name of the method in
which it is being defined, and spits out a visual representation of
the state machine.

Examples:
```
$ dump_state_machine tensorpipe/core/pipe_impl.cc --method=advanceReadOperation -- -I. | dot -Tpng > pipe_read.png
```
pipe_read

```
$ dump_state_machine tensorpipe/core/pipe_impl.cc --method=advanceWriteOperation -- -I. | dot -Tpng > pipe_read.png
```
pipe_write

### DIFF
--- a/tensorpipe/misc/dump_state_machine.cc
+++ b/tensorpipe/misc/dump_state_machine.cc
@@ -1,0 +1,126 @@
+#include <iostream>
+#include <regex>
+#include <unordered_set>
+
+#include <clang/ASTMatchers/ASTMatchFinder.h>
+#include <clang/ASTMatchers/ASTMatchers.h>
+#include <clang/Frontend/FrontendActions.h>
+#include <clang/Tooling/CommonOptionsParser.h>
+#include <clang/Tooling/Tooling.h>
+#include <llvm/Support/CommandLine.h>
+
+using namespace clang::ast_matchers;
+using namespace clang::tooling;
+using namespace llvm;
+
+std::string exprToString(const clang::Expr& e) {
+  std::string statement;
+  raw_string_ostream stream(statement);
+  e.printPretty(stream, nullptr, clang::PrintingPolicy(clang::LangOptions()));
+  stream.flush();
+
+  return statement;
+}
+
+std::string cleanUp(const std::string& s) {
+  std::string res = s;
+  res = std::regex_replace(res, std::regex("(struct|class) [a-zA-Z_]+::"), "");
+  res = std::regex_replace(res, std::regex("this->"), "");
+  return res;
+}
+
+std::string escape(const std::string& s) {
+  std::string res = s;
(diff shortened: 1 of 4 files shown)


tensorpipe/misc/dump_state_machine.cc
```diff
@@ -0,0 +1,126 @@
...
+    int edgeWeight = 100;
+    std::string cond = cleanUp(exprToString(*e.getArg(3)));
+    if (std::regex_search(cond, std::regex("^error_"))) {
+      edgeColor = "red3";
+      edgeWeight = 0;
+    }
+    if (std::regex_search(cond, std::regex("^!error_"))) {
+      edgeColor = "forestgreen";
+    }
+    cond = std::regex_replace(cond, std::regex(" \\&\\&"), "\\n");
```
lw, line 78: This means it won't handle more complicated expressions like `A || (B && C)` right? I think it's fine for now, we'll fix if if we'll ever need to.

tensorpipe/misc/dump_state_machine.cc
```diff
@@ -0,0 +1,126 @@
...
+ public:
+  virtual void run(const MatchFinder::MatchResult& Result) {
+    static int edgeCount = 0;
+
+    const clang::CallExpr& e = *Result.Nodes.getNodeAs<clang::CallExpr>("x");
+    std::string edgeId = "edge" + std::to_string(edgeCount++);
+    std::string fromId = cleanUp(exprToString(*e.getArg(1)));
+    std::string toId = cleanUp(exprToString(*e.getArg(2)));
+
+    if (nodes_.count(fromId) == 0) {
```
lw, line 60: Nit: this check could be done inside `addNode`.

lw (approved):
I must say, I was expecting to find some Python-based regex code (or, worse, Bash-based or Perl-based), but I'm surprised to see proper AST parsing using clang! Good stuff!!

I know nothing of Clang, nor of the dot format, hence I can't really weigh in on that. However the code definitely works, given those amazing graphs you put in the summary!

I wonder if we could consider using that dot format (or perhaps some other intermediate format) to do some validation/linting of the state machines, like checking that for each node the conditions of all outgoing edges partition the space of all options. Or the fact that if the action name contains `DOING_FOO` then for each edge either the target state still contains `DOING_FOO` or the condition contains `doneDoingFoo`. Stuff like this.

tensorpipe/misc/dump_state_machine.cc
beauby: That's correct.

PR #373, opened 2021-04-26
Document support for Linux features

lw:
Differential Revision: D28000214

### DIFF
--- a/docs/linux_support.md
+++ b/docs/linux_support.md
@@ -1,0 +1,95 @@
+This document is intended for developers and advanced users. It’s the kind of document that risks going out of date very quickly, hence take it with a grain of salt.
+
+In order to try to be as performant as possible, TensorPipe sometimes relies on new and advanced kernel features. This is causing issues to users who are building and/or running on old kernels. Hence, whenever we use such features, we should always “guard” them somehow, i.e., detect their availability at compile-time or (preferably) at runtime, and disable the backend or mark it non-viable. It is ok-ish for users with old kernels to not have access to all backends, as long as there’s always at least one backend they can use.
+
+## Compile-time vs runtime, Linux vs glibc
+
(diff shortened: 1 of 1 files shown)


docs/linux_support.md
```diff
@@ -0,0 +1,95 @@
...
+
+An additional complication is added by the fact that we typically access syscalls through their glibc wrappers. First of all, this means we only get access to a syscall once glibc wraps it, which could happen years later. But it also means we link to a glibc symbol, and thus to a specific version of glibc’s shared object. With the kernel, using an unsupported feature results in a runtime error when first used, which we can catch; but with glibc we get a linker error due to missing symbols at startup, even if the user doesn’t use TensorPipe, even if we could “tolerate” these symbols’ absence. It is thus desirable at times to avoid the glibc wrappers.
```
beauby, line 9:
```suggestion
An additional complication is added by the fact that we typically access syscalls through their glibc wrappers. First of all, this means we only get access to a syscall once glibc wraps it, which could happen years later. But it also means we link to a glibc symbol, and thus to a specific version of glibc’s shared object. With the kernel, using an unsupported feature results in a runtime error when first used, which we can catch; but with glibc we get a loader error due to missing symbols at startup, even if the user doesn’t use TensorPipe, even if we could “tolerate” these symbols’ absence. It is thus desirable at times to avoid the glibc wrappers.
```

docs/linux_support.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Linux 2.3.16 (September 1999)
+
+* The `/proc/sys/kernel/random/boot_id` file. See `random(4)`.
+
+  No git hash as it predates the use of git by Linux[https://github.com/torvalds/linux/blob/1da177e4c3f41524e886b7f1b8a0c1fc7321cac2/drivers/char/random.c#L1270-L1278](https://github.com/torvalds/linux/blob/1da177e4c3f41524e886b7f1b8a0c1fc7321cac2/drivers/char/random.c#L1270-L1278)
+
+### Linux 2.6.26 (July 2008)
+
+* Version 3 of Linux capabilities. (Initial capability support, including the `capget` syscall, dates back to Linux 2.1.100, from May 1998). See `capget(2)`.
```
beauby, line 34:
```suggestion
* Version 3 of Linux capabilities. (Initial capability support, including the `capget` syscall, dates back to Linux 2.1.100, from May 1998.) See `capget(2)`.
```

beauby (approved):

beauby:
LGTM

PR #378, opened 2021-05-02
Factor out channel selection.

beauby:
In order to make channel selection testable, we need to move it out of
the anonymous namespace in which it is currently being declared.

The `SelectedChannels` struct is being renamed to `ChannelSelection`, otherwise the code is simply moved from `core/pipe_impl.cc` to `core/channel_selection.cc`.

### DIFF
--- a/tensorpipe/test/core/channel_selection_test.cc
+++ b/tensorpipe/test/core/channel_selection_test.cc
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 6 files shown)


tensorpipe/test/core/channel_selection_test.cc
```diff
@@ -0,0 +1,151 @@
...
+} // namespace
+
+TEST(ChannelSelection, NoCompatibleChannel) {
+  auto channelContext = MockChannelContext::create({
+      {Device{kCpuDeviceType, 0}, "any"},
+  });
+
+  ChannelSelection channelSelection = selectChannels(
+      {
+          {0, std::make_tuple("foo_channel", channelContext)},
```
beauby, line 72: The `make_tuple` here is required (only) by gcc5.

PR #380, opened 2021-05-03
Add tests for channel::Context::canCommunicateWithRemote.

beauby:
* #384 CudaXth: copy from worker thread.
* #383 Add support for CPU-to-GPU/GPU-to-CPU in CudaXth.
* #381 Remove CPU-to-CPU test from XDTT tests.
* **#380 Add tests for channel::Context::canCommunicateWithRemote.**

Differential Revision: [D28255095](https://our.internmc.facebook.com/intern/diff/D28255095)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,32 +303,32 @@ workflows:
           cuda_version: "9.2"
           # Excluding CudaGdr for lack of InfiniBand hardware, and CudaIpc on
           # multi GPU for lack of p2p capabilities.
-          exclude_tests: "CudaGdr*:CudaIpc/CudaMultiGPUChannelTestSuite*"
+          exclude_tests: "CudaGdr*:CudaIpc/CudaMultiGPUChannelTestSuite*:*CudaIpcCanCommunicateWithRemote*"
       - build_gpu:
           name: GPU (CUDA 10.1)
           cuda_version: "10.1"
           # Excluding CudaGdr for lack of InfiniBand hardware, and CudaIpc on
           # multi GPU for lack of p2p capabilities.
-          exclude_tests: "CudaGdr*:CudaIpc/CudaMultiGPUChannelTestSuite*"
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -303,32 +303,32 @@ workflows:
           cuda_version: "9.2"
           # Excluding CudaGdr for lack of InfiniBand hardware, and CudaIpc on
           # multi GPU for lack of p2p capabilities.
-          exclude_tests: "CudaGdr*:CudaIpc/CudaMultiGPUChannelTestSuite*"
+          exclude_tests: "CudaGdr*:CudaIpc/CudaMultiGPUChannelTestSuite*:*CudaIpcCanCommunicateWithRemote*"
```
lw, line 306: Is there a way we could actually test IPC too? For example, if we set CUDA_VISIBLE_DEVICES=0 on both endpoints, they would see the same device and this device would support P2P onto itself, thus IPC should be viable.

lw (approved):
It seems in most cases (i.e., all except CUDA basic) we just check if all pairs can communicate. I wonder if we can be a bit more specific than this?

Like, could we add a check that, for CPU channels, the device descriptors contain at least one key that is CPU? And same for CUDA channels? Could we check that, for CUDA channels, there's one descriptor for each visible device?

Ideally I'd also like to check "negative" responses, e.g., ensure that XTH returns false when used across-process, etc. However this will take a lot more effort, as it will require a specific test setup for each channel, kind of like what we do for CMA today already.

beauby:
@beauby has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D28255095).

beauby:
@beauby has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D28255095).

PR #381, opened 2021-05-03
Remove CPU-to-CPU test from XDTT tests.

beauby:
* #384 CudaXth: copy from worker thread.
* #383 Add support for CPU-to-GPU/GPU-to-CPU in CudaXth.
* **#381 Remove CPU-to-CPU test from XDTT tests.**
* #380 Add tests for channel::Context::canCommunicateWithRemote.

As we do not expect CUDA channels to be used for CPU-to-CPU transfers,
let's remove this test (which would fail for a channel that does not
actually support it, like the coming CudaXth).

Differential Revision: [D28255208](https://our.internmc.facebook.com/intern/diff/D28255208)


lw (approved):
I guess this is because CUDA basic doesn't support CPU-to-CPU anymore?

beauby:
@beauby has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D28255208).

PR #391, opened 2021-06-06
Cmake files restructuring changes

pruthvistony:


pruthvistony:
@lw @jeffdaily @jithunnair-amd,
Please review the changes.

tensorpipe/CMakeLists.txt
```diff
@@ -4,8 +4,16 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-add_library(tensorpipe
-  ${TP_STATIC_OR_SHARED}
+# List of source files that we need to build libtensorpipe.
+set(TP_PUBLIC_SRCS)
+set(TP_PRIVATE_SRCS)
+set(TP_CUDA_PRIVATE_SRCS)
```
lw, line 10: As I mentioned in the other PR, I don't see the reason to distinguish public vs private sources. They both end up in the same library. Are there downsides to merging them and just using two variables (`TP_SRCS` and `TP_CUDA_SRCS`)? (Note that headers would still need to be separated between public and private)

tensorpipe/CMakeLists.txt
```diff
@@ -4,8 +4,16 @@
...
+
+# List of header files that we need to install.
+set(TP_PUBLIC_HDRS)
+set(TP_CUDA_HDRS)
```
lw, line 14: Can we call it `TP_PUBLIC_CUDA_HDRS`?

I realize that _right now_ it looks silly to define public headers without defining the private ones, but internally we need to define the private ones too and I would like to keep the definitions aligned. Ideally one day we should find a way to have a single source of truth for both the internal and CMake build systems...

tensorpipe/CMakeLists.txt
```diff
@@ -206,29 +202,56 @@ if(TP_ENABLE_IBV)
...
-  target_link_libraries(tensorpipe PRIVATE ${CF} ${IOKIT})
+# Handling the file lists
+if(TP_USE_CUDA)
+  list(APPEND TP_PRIVATE_SRCS ${TP_CUDA_PRIVATE_SRCS})
+  list(APPEND TP_PUBLIC_HDRS ${TP_CUDA_HDRS})
 endif()
 
+# Add the library target
+add_library(tensorpipe ${TP_STATIC_OR_SHARED} ${TP_PUBLIC_SRCS})
+target_sources(tensorpipe PRIVATE ${TP_PRIVATE_SRCS})
```
lw, line 220: As mentioned above I'm convinced these two can be joined. The former distinction was probably some cruft.

tensorpipe/CMakeLists.txt
```diff
@@ -206,29 +202,56 @@ if(TP_ENABLE_IBV)
...
+  target_link_libraries(tensorpipe PUBLIC ${CUDA_LIBRARIES})
+  target_include_directories(tensorpipe PUBLIC ${CUDA_INCLUDE_DIRS})
+endif()
+
+# MAC OS specific library deps
+if(APPLE)
+  find_library(CF CoreFoundation)
+  find_library(IOKIT IOKit)
+  target_link_libraries(tensorpipe PRIVATE ${CF} ${IOKIT})
+endif()
```
lw, line 243: As I said in the other PR, it could be nicer to define some additional variables at the beginning, like `TP_LINK_LIBRARIES` and `TP_INCLUDE_DIRECTORIES`, ..., and then append elements to them as we go (so that we can keep this logic more tightly coupled with the rest and avoid repetition), and at the end just use those variables to populate the `tensorpipe` target.

tensorpipe/test/CMakeLists.txt
```diff
@@ -4,7 +4,13 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-add_executable(tensorpipe_test
+# List of source files that we need to build tensorpipe_test executable.
+set(TP_TEST_PUBLIC_SRCS)
+set(TP_TEST_PRIVATE_SRCS)
+set(TP_TEST_CUDA_PRIVATE_SRCS)
+set(TP_TEST_CUDA_KERNEL_SRCS)
```
lw, line 11: Like before: please don't separate public-vs-private sources.

tensorpipe/test/CMakeLists.txt
```diff
@@ -50,42 +56,50 @@ if(TP_ENABLE_IBV)
...
 if(TP_ENABLE_CMA)
-  target_sources(tensorpipe_test PRIVATE
+  list(APPEND TP_TEST_PRIVATE_SRCS
     channel/cma/cma_test.cc
     )
   add_subdirectory(channel/cma)
 endif()
 
+add_executable(tensorpipe_test ${TP_TEST_PUBLIC_SRCS})
+target_sources(tensorpipe_test PRIVATE ${TP_TEST_PRIVATE_SRCS})
```
lw, line 66: Why define the target before we've even finished determining its sources? This seems to defeat the purpose of the `TP_TEST_..._SRCS` variables, as in, if we're going to incrementally add sources to the target then we could just define the target at the beginning and just drop those variables.

lw:
One thing to add: we'll very soon split the CUDA-related parts of TensorPipe in a separate "add-on" library, so that we will have both a `tensorpipe` and a `tensorpipe_cuda` target, the former one including all the core classes and the second one depending on the former one. I don't expect this to cause any problems, and I'll fix the CMake files to make it work. I think this might potentially make adding ROCm support even easier? (Since we wouldn't need to touch at all the `tensorpipe` target, and just the `tensorpipe_cuda` one).

tensorpipe/CMakeLists.txt
pruthvistony: I have updated the changes. Was thinking of any symbol conflicts which would arise with PUBLIC.

tensorpipe/CMakeLists.txt
pruthvistony: done

tensorpipe/CMakeLists.txt
pruthvistony: done.

tensorpipe/CMakeLists.txt
pruthvistony: Updated the changes.
Keeping all the dependent libraries link as PRIVATE, so that the interface function dont become public API of tensorpipe library. And keeping the target include directories as PUBLIC, since it is required.

tensorpipe/test/CMakeLists.txt
pruthvistony: done

tensorpipe/test/CMakeLists.txt
pruthvistony: We can do this way too, but I feel we have a cleaner way now with the list of files and added to `add_executable() `and `target_sources()` removed.

pruthvistony:
> One thing to add: we'll very soon split the CUDA-related parts of TensorPipe in a separate "add-on" library, so that we will have both a `tensorpipe` and a `tensorpipe_cuda` target, the former one including all the core classes and the second one depending on the former one. I don't expect this to cause any problems, and I'll fix the CMake files to make it work. I think this might potentially make adding ROCm support even easier? (Since we wouldn't need to touch at all the `tensorpipe` target, and just the `tensorpipe_cuda` one).

Yes, I agree this will be a good change. I could put the ROCm (HIP) related files into `tensorpipe_rocm`.

jithunnair-amd:
> > One thing to add: we'll very soon split the CUDA-related parts of TensorPipe in a separate "add-on" library, so that we will have both a `tensorpipe` and a `tensorpipe_cuda` target, the former one including all the core classes and the second one depending on the former one. I don't expect this to cause any problems, and I'll fix the CMake files to make it work. I think this might potentially make adding ROCm support even easier? (Since we wouldn't need to touch at all the `tensorpipe` target, and just the `tensorpipe_cuda` one).
> 
> Yes, I agree this will be a good change. I could put the ROCm (HIP) related files into `tensorpipe_rocm`.

Actually, if we follow the pytorch pattern, it should be `tensorpipe_hip`. But I'd like Luca to confirm if he was intending the name of the target to be different for CUDA vs ROCm (that'd be our preference to align with how Pytorch/Gloo etc. do it).

lw:
I don't have strong opinions on calling the hipified version `tensorpipe_hip`. However, if I were to encounter such a library, my instinct would be to assume that it's distinct from `tensorpipe_cuda`, providing different symbols, and thus wouldn't conflict with it. However, it's not the case (yet?) because given how we currently intend to hipify TensorPipe they would both contain a CUDABuffer symbol, and the channels would be named the same, ... Thus I'm perhaps more inclined to keep calling the hipified version `tensorpipe_cuda`, and perhaps change it if we'll change our "hipification style".

pruthvistony:
@lw 
Can you please let me know more details of the failing test, if it is related to this change.

Test(s) failing

Test(s) failing internally on Facebook.
failed:
fbsource-target-determinator

lw:
Don't worry, I'll figure out the internal stuff, though I'm sure it's just a flake since the CMake files are completely ignored internally...

I hope to land this change by tomorrow.

PR #398, opened 2021-06-24
[ROCm] Hipify changes

pruthvistony:
- Add Hipify as a git submodule
- Trigger hipify from cmake build
- TP_USE_ROCM controls the trigger, which will be set to ON
  when building on ROCm

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,3 +11,6 @@
 [submodule "third_party/libnop"]
 	path = third_party/libnop
 	url = https://github.com/google/libnop.git
+[submodule "third_party/hipify"]
+	path = third_party/hipify
+	url = https://github.com/ROCmSoftwarePlatform/hipify-torch.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,11 @@ include(Sanitize)
 # Misc checks to cope with various compiler modes.
 include(MiscCheck)
 
+# ROCm related
+if (TP_USE_ROCM)
+  include(Hipify)
+endif()
+
 add_subdirectory(tensorpipe)
 
 install(EXPORT TensorpipeTargets
--- a/cmake/Options.cmake
+++ b/cmake/Options.cmake
@@ -31,6 +31,12 @@ endmacro()
 
 # TODO: Default to ON if CUDA available.
 option(TP_USE_CUDA "Enable support for CUDA tensors" OFF)
+option(TP_USE_ROCM "Enable support for ROCM tensors" OFF)
+
+# if both TP_USE_CUDA and TP_USE_ROCM is set then break
+if(TP_USE_CUDA AND TP_USE_ROCM)
+  message(FATAL_ERROR "Tensorpipe can be built either for CUDA or ROCm, TP_USE_CUDA and TP_USE_ROCM both are set, erroring out!!!!")
+endif()
 
 # Optional features
 option(TP_BUILD_BENCHMARK "Build benchmarks" OFF)
--- a/tools/amd_build/build_amd.py
+++ b/tools/amd_build/build_amd.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import os
+import sys
+import argparse
+
+sys.path.append(os.path.realpath(os.path.join(
+    os.path.dirname(__file__),
+    os.path.pardir,
+    os.path.pardir,
+    'third_party')))
+
+from hipify import hipify_python
+
+parser = argparse.ArgumentParser(description='Top-level script for HIPifying, filling in most common parameters')
+parser.add_argument(
+    '--project-directory',
+    type=str,
+    help="The root of the project. (default: %(default)s)",
+    required=True)
+
+parser.add_argument(
+    '--output-directory',
+    type=str,
+    help="The Directory to Store the Hipified Project",
+    required=True)
+
+args = parser.parse_args()
+
+includes = [
+    '*'
+]
+
+ignores = [
+]
+
+# capturing the return value which is a dict[filename]:HipifyResult
+HipifyFinalResult = hipify_python.hipify(
+    project_directory=args.project_directory,
+    output_directory=args.output_directory,
+    includes=includes,
+    ignores=ignores,
+    is_pytorch_extension=True)
--- a/cmake/Hipify.cmake
+++ b/cmake/Hipify.cmake
@@ -1,0 +1,23 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# cmake file to trigger hipify
+
+set(HIPIFY_SCRIPTS_DIR ${PROJECT_SOURCE_DIR}/tools/amd_build)
+set(HIPIFY_COMMAND
+  ${HIPIFY_SCRIPTS_DIR}/build_amd.py
+  --project-directory ${PROJECT_SOURCE_DIR}
+  --output-directory ${PROJECT_SOURCE_DIR}
+)
+
+execute_process(
+  COMMAND ${HIPIFY_COMMAND}
+  RESULT_VARIABLE hipify_return_value
+)
+if (NOT hipify_return_value EQUAL 0)
+  message(FATAL_ERROR "Failed to hipify files!")
+endif()
+


pruthvistony:
@lw @jeffdaily, @jithunnair-amd,
Please review the changes.

jithunnair-amd:
@pruthvistony I think we discussed this before, but just to make sure: could the build_amd.py be part of hipify-torch so that it doesn't have to be added to the hipifying project's sources? Is there anything that's project-specific that cannot be passed in as arguments to build_amd.py?

pruthvistony:
> @pruthvistony I think we discussed this before, but just to make sure: could the build_amd.py be part of hipify-torch so that it doesn't have to be added to the hipifying project's sources? Is there anything that's project-specific that cannot be passed in as arguments to build_amd.py?

As discussed, build_amd.py can be moved to hipify-torch repo and triggered directly from cmake. Just that passing all parameters like list(regex) can be tricky within cmake. Currently I dont see anything specific problem as such in passing arguments through build_amd.py. I have kept current change same as it is used in other projects. Can update this change as an improvement, since it is not a blocker for other ROCm build related changes.

tools/amd_build/build_amd.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
```
lw, line 1: We have been putting some assorted tools in the `tensorpipe/misc` folder. It would be good to use a single directory for all these.

tools/amd_build/build_amd.py
```diff
@@ -0,0 +1,49 @@
...
+
+import os
+import sys
+import argparse
+
+sys.path.append(os.path.realpath(os.path.join(
+    os.path.dirname(__file__),
+    os.path.pardir,
+    os.path.pardir,
+    'third_party')))
```
lw, line 17: This could work, though it means that this file "assumes" a certain structure in the repo, which means it's not really standalone and thus possibly brittle. I now wonder whether it could be more robust to have the path of the `hipify` module passed in via a command-line argument?

tools/amd_build/build_amd.py
```diff
@@ -0,0 +1,49 @@
...
+ignores = [
+]
+
+# capturing the return value which is a dict[filename]:HipifyResult
+HipifyFinalResult = hipify_python.hipify(
+    project_directory=args.project_directory,
+    output_directory=args.output_directory,
+    includes=includes,
+    ignores=ignores,
+    is_pytorch_extension=True)
```
lw, line 49: What does the `is_pytorch_extension` flag do? TensorPipe is not a PyTorch extension.

tools/amd_build/build_amd.py
```diff
@@ -0,0 +1,49 @@
...
+
+sys.path.append(os.path.realpath(os.path.join(
+    os.path.dirname(__file__),
+    os.path.pardir,
+    os.path.pardir,
+    'third_party')))
+
+from hipify import hipify_python
+
+parser = argparse.ArgumentParser(description='Top-level script for HIPifying, filling in most common parameters')
```
lw, line 21: It is generally more idiomatic to put all the logic of a runnable Python entry-point script in a `def main():` function and then invoke it with `if __name__ == "__main__": main()`.

lw (commented):
I could be OK with merging this but (correct me if I'm wrong) I don't think this is yet usable is it? If anyone tried to run with `TP_USE_ROCM` the build would most likely fail. Should we fix all that and ensure everything works before we do that?

Also, while I guess we cannot run any ROCm tests on CircleCI (because it's lacking the necessary hardware), I think we should be able to at least add a ROCm build job. Do you think you could do that? Thanks!

cmake/Options.cmake
```diff
@@ -31,6 +31,12 @@ endmacro()
 
 # TODO: Default to ON if CUDA available.
 option(TP_USE_CUDA "Enable support for CUDA tensors" OFF)
+option(TP_USE_ROCM "Enable support for ROCM tensors" OFF)
+
+# if both TP_USE_CUDA and TP_USE_ROCM is set then break
+if(TP_USE_CUDA AND TP_USE_ROCM)
+  message(FATAL_ERROR "Tensorpipe can be built either for CUDA or ROCm, TP_USE_CUDA and TP_USE_ROCM both are set, erroring out!!!!")
+endif()
```
beauby, line 39: I am not suggesting to do this now, but how difficult would that limitation be to lift?

lw: For the first version of ROCm support we intend to still use the `CudaBuffer` name (and the `cuda_ipc`, ... ones) for the HIP version of TensorPipe as well, out of simplicity, hence using the CUDA and HIP versions together would cause name conflicts. Eventually I think we could/should fix this.

pruthvistony:
> I could be OK with merging this but (correct me if I'm wrong) I don't think this is yet usable is it? If anyone tried to run with `TP_USE_ROCM` the build would most likely fail. Should we fix all that and ensure everything works before we do that?
> 
> Also, while I guess we cannot run any ROCm tests on CircleCI (because it's lacking the necessary hardware), I think we should be able to at least add a ROCm build job. Do you think you could do that? Thanks!

Currently pyTorch ROCm tests are executed on jenkins. 
The changes are not yet usable, I am raising changes in multiple PR as it was suggested previously.  Next PR which builds tensorpipe enabling 'TP_USE_ROCM' is in draft mode, but is not complete due to a not support API. For which we are working with the HIP team.

Regarding the ROCm build, I believe job can be setup. Regarding the tests I will check on the necessary hardware and get back.

tools/amd_build/build_amd.py
pruthvistony: In the process of moving this file into hipify-torch repo, which is added a git submodule. 
https://github.com/ROCmSoftwarePlatform/hipify-torch.git
Once the PRs into hipify-torch is merged, I will update this PR.

tools/amd_build/build_amd.py
pruthvistony: Since this file will be moved hipify-torch repo, the above code will not be required.

tools/amd_build/build_amd.py
pruthvistony: Yes, tensorpipe is not a extension, but this parameter is used internal in hipify.
I will let @jithunnair-amd to comment for info here.

cmake/Options.cmake
pruthvistony: We are thinking tensorpipe build for CUDA or ROCm will be mutual exclusive, since pyTorch lib is also mutual exclusive.
Even for channels there will be only hip based channels, `hip_basic, hip_gdr, hip_ipc`, ... . 
So the builds will have corresponding flags turned ON. So I didn't get whats the concern here. Please let me know more details.

tools/amd_build/build_amd.py
jithunnair-amd: That nomenclature made sense until now :) Essentially, we needed a way to preserve the "legacy" way of hipifying for PyTorch source code, but use the "new and improved" way for other code which used hipify, which were basically PyTorch extensions so far. We'll assess the best way to generalize this, but this flag is needed for now until that happens.

lw: Ok, got it, just wanted to check. It would be nice if we could add a comment to explain this, so I won't wonder about it again the next time I read this code. (Though if this code is moved to the hipify repo it won't matter anyways)

cmake/Options.cmake
lw: I think the question here was motivated by the fact that in principle there shouldn't be any hard blocker to have both CUDA and ROCm (once we fix the name conflicts) right? Hence this is mainly a comment about "code style". In practice though yes, when used from PyTorch they will be mutually exclusive, hence no worries about this.

lw:
> Currently pyTorch ROCm tests are executed on jenkins.

Yes, I saw that. However that's because they are both build and _run_ on AMD GPUs right? If we only built, without running, we could do so on machines without any GPUs. (PyTorch already builds the CUDA version on CPU-only machines). It wouldn't be perfect but it would at least catch build issues, for example caused by hipification.

pruthvistony:
Moved all the hipify related files into the hipify-torch [repo](https://github.com/ROCmSoftwarePlatform/hipify-torch).

tools/amd_build/build_amd.py
pruthvistony: Updated the code as suggest and committed to hipify [repo](https://github.com/ROCmSoftwarePlatform/hipify-torch)

CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ project(tensorpipe LANGUAGES C CXX)
 
 set(CMAKE_CXX_STANDARD 14)
 
-list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" "${PROJECT_SOURCE_DIR}/third_party/hipify/cmake")
```
beauby, line 13: Please move the `list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/third_party/hipify/cmake")` to the `if (TP_USE_ROCM)` section, so that everything is in one place.

.gitmodules
```diff
@@ -11,3 +11,6 @@
 [submodule "third_party/libnop"]
 	path = third_party/libnop
 	url = https://github.com/google/libnop.git
+[submodule "third_party/hipify"]
```
mrshenli, line 14: OOC, this dependency does not seem existing in PyTorch's `.gitmodules`. Any reason for this difference?

CMakeLists.txt
pruthvistony: Updated as suggested.

.gitmodules
jithunnair-amd: Yes, that's because PyTorch currently has hipify logic as part of its source code at https://github.com/pytorch/pytorch/blob/master/torch/utils/hipify
We can't use that for Tensorpipe as it'll create a circular dependency.

pruthvistony: Yes, in PyTorch hipify is not used as a git submodule. For new project to hipify we are using hipify-torch repo as a git submodule, so that all the hipification code is in a single place, providing many interfaces.

mrshenli: Do we plan to change PyTorch's hipify to use the same strategy?

jithunnair-amd: Yes, we would like to in the long run, but PyTorch being a much bigger codebase, and having a more complex hipification strategy, will require much more coordination and effort.

PR #409, opened 2021-09-03
EFA implementation 

VoVAllen:
TODO:
- Add logs
- Handle the error in the write way (Shall I use setError or directly throw errow?)
- Make libfabirc dependency optional


VoVAllen:
Explanation for the implementation: 
https://github.com/pytorch/tensorpipe/blob/f9b14e57e7e4c4c818ca9b402438a6a307729cd9/tensorpipe/transport/efa/mode.md

PR #435, opened 2022-03-25
CMake: support find_package

### LINKED ISSUE #320: CMake: config file is not installed along targets file targets file
Installing `TensorpipeTargets.cmake` and `TensorpipeTargets-<mode>.cmake` is useless if a config file including `TensorpipeTargets.cmake` is not installed with these files: https://cmake.org/cmake/help/latest/command/find_package.html#full-signature-and-config-mode

You might also want to export tensorpipe target with a namespace to follow best practices (and add an alias target to be consistent between `add_subdirectory()` and `find_package()`).

luncliff:
Resolves #320 

### Changes

* Create TensorpipeConfig.cmake from the template
* Use CMAKE_INSTALL_DATAROOTDIR for share/Tensorpipe
* Add `pytensorpipe` to install targets

```cmake
find_package(Tensorpipe CONFIG REQUIRED)
```

```console
$ cmake -Bbuild -DTensorpipe_DIR=/path/to/install/share/cmake/Tensorpipe # ...
```

Install result should be like this.

```console
$ tree .
.
├── include
│   └── tensorpipe
├── lib
│   ├── libtensorpipe.so
│   ├── libtensorpipe_cuda.so
│   ├── libtensorpipe_uv.a
│   └── pytensorpipe.cpython-38-x86_64-linux-gnu.so
└── share
    └── cmake
        └── Tensorpipe
            ├── TensorpipeConfig.cmake
            ├── TensorpipeTargets-debug.cmake
            └── TensorpipeTargets.cmake

``` 

### Related Works

* https://github.com/microsoft/vcpkg/pull/23569

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,5 +24,16 @@ include(MiscCheck)
 add_subdirectory(tensorpipe)
 
 install(EXPORT TensorpipeTargets
-        DESTINATION share/cmake/Tensorpipe
+        DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/cmake/Tensorpipe
         FILE TensorpipeTargets.cmake)
+
+# Create TensorpipeConfig.cmake for find_package(Tensorpipe CONFIG)
+include(CMakePackageConfigHelpers)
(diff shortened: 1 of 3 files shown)


talregev:
Any news? why the tests are failing? Should we run them again?

